PR #147, opened 2023-07-13
Add proteus module for maxquant data analysis

WackerO:
This PR adds the proteus module to differentialabundance; it allows importing proteomics measurements from MaxQuant which can then be analyzed with the limma module

## PR checklist

- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [x] Output Documentation in `docs/output.md` is updated.


pinin4fjords (changes requested):
A couple of initial comments while I look at the rest.

pinin4fjords (changes requested):
I'm having a bit of trouble in this, in that you're duplicating a lot of code from the wider pipeline. We should be able to process conditionally in certain places, e.g. in how the matrices are generated which differential method is called, but basically follow the same structure. 

e.g. CUSTOM_MATRIXFILTER should only really need to be called in one place. It should be possible to validate everything the same way, since we need to check the same things- i.e. that samples, matrices and metadata are compatible.

We should be able to handle this similarly to arrays. I'll take a look and see if I can make some more detailed suggestions, but maybe that is enough to start with.

workflows/differentialabundance.nf
```diff
@@ -113,149 +125,219 @@ workflow DIFFERENTIALABUNDANCE {
     // Set up some basic variables
     ch_versions = Channel.empty()
 
-    // If we have affy array data in the form of CEL files we'll be deriving
-    // matrix and annotation from them
-
-    if (params.study_type == 'affy_array'){
+    // Check first if maxquant, as in that case the entire VALIDATE block can be skipped
```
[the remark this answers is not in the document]
WackerO: Indeed, but the maxquant matrices have a slightly different structure compared to the rnaseq stuff so the validator does not work for them ATM. I can check how it needs to be changed so it can be kept for the Px path as well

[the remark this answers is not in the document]
WackerO: Okay, the validator worked actually with the proteus output. The checks are now kind of double but nvm

workflows/differentialabundance.nf
```diff
@@ -157,8 +171,44 @@ workflow DIFFERENTIALABUNDANCE {
         ch_versions = ch_versions
             .mix(AFFY_JUSTRMA_RAW.out.versions)
 
-    }
-    else if(params.study_type == 'geo_soft_file'){
+    } else if (params.study_type == 'maxquant'){
+        // For maxquant, run proteus module to import the protein abundances
+
+        // Save contrasts file to channel
+        ch_contrasts_file = Channel.from([[exp_meta, file(params.contrasts)]])
```
pinin4fjords, line 178: This is already defined below the conditional - just relocate that to above the conditional so you can use it here

workflows/differentialabundance.nf
```diff
@@ -157,8 +171,44 @@ workflow DIFFERENTIALABUNDANCE {
...
+
+        // For the plotting and following modules, save all contrast info
+        ch_contrasts = ch_contrasts_split
+            .map{
+                it.blocking = it.blocking.replace('NA', '')
+                if (!it.id){
+                    it.id = it.values().join('_')
+                }
+                tuple(it, it.variable, it.reference, it.target)
+            }
```
pinin4fjords, line 202:
```suggestion
```

Do you need any of this? `ch_contrasts` is redefined before LIMMA anyway, based on the VALIDATOR output

workflows/differentialabundance.nf
```diff
@@ -157,8 +171,44 @@ workflow DIFFERENTIALABUNDANCE {
...
+            .map{ it.tail().first() }
+
+        // For proteus, extract only meta and contrast variable
+        ch_contrasts_proteus = ch_contrasts_split
+            .map{
+                tuple(
+                    exp_meta,       // meta map
+                    it.variable     // contrast variable
+                )
+            }
```
pinin4fjords, line 192:
```suggestion
        ch_contrasts_split = ch_contrasts_file
            .splitCsv ( header:true, sep:(params.contrasts.endsWith('tsv') ? '\t' : ','))
            .map{ it.tail().first() }

        // For proteus, extract only meta and contrast variable
        ch_contrasts_proteus = ch_contrasts_split
            .map{
                tuple(
                    exp_meta,       // meta map
                    it.variable     // contrast variable
                )
            }
```

So, proteus will run once for every contrast variable? How does the contrast variable impact on the quantifications?

If it does, why does the proteus module return results keyed by the `meta`, rather than the `meta2` corresponding to the contrast variable?

I think we can simplify this chunk a lot, but I need to understand what's going on better first.

workflows/differentialabundance.nf
WackerO: Afraid so; PROTEUS needs the contrast variable, and in turn, the VALIDATOR needs PROTEUS output

pinin4fjords: I think you're missing something- this definition of `ch_contrasts` is never used. It's redefined [here](https://github.com/nf-core/differentialabundance/blob/4d876af4756a5b2822fc31235ef16021e8db77c6/workflows/differentialabundance.nf#L325).

workflows/differentialabundance.nf
```diff
@@ -28,6 +28,18 @@ if (params.study_type == 'affy_array'){
     } else {
         error("CEL files archive not specified!")
     }
+} else if (params.study_type == 'maxquant') {
+    
+        // Should the user have enabled --shinyngs_build_app and/or --gsea_run, throw an error
+        if (params.shinyngs_build_app) {
+            error("Cannot build shinyngs app for maxquant data; please set --shinyngs_build_app to false.")
```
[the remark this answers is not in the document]
WackerO: The NAs are the blocker :*D If an na.omit is added to the shiny code or some additional filtering done before (maybe even with some super fancy nextflow code?), it should work without problems, but I'd like to move such changes to a future PR

workflows/differentialabundance.nf
WackerO: You are right...I removed that

assets/differentialabundance_report.Rmd
```diff
@@ -232,22 +233,73 @@ assay_names <- simpleSplit(params$exploratory_assay_names)
...
+# Set up vector of unlogged assay files (if any)
+unlogged <- c()
+features_log2_assays <- params$features_log2_assays
+
+# For maxquant input override param as all assays are already logged by proteus
+if (params$study_type == 'maxquant') {
+  features_log2_assays <- ""
+}
+
+if (is.null(features_log2_assays)) {
```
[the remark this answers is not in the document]
pinin4fjords: @WackerO you resolved this, but you're still repeating code here rather than using the factored-out functionality. To be clear, I meant [this function](https://github.com/pinin4fjords/shinyngs/blob/b544d858e0c64879b12b5542c3e7cd0141d939a0/R/accessory.R#L1291). You can it being used in the exploratory script [here](https://github.com/pinin4fjords/shinyngs/blob/b544d858e0c64879b12b5542c3e7cd0141d939a0/exec/exploratory_plots.R#L189).

workflows/differentialabundance.nf
```diff
@@ -174,6 +221,15 @@ workflow DIFFERENTIALABUNDANCE {
         ch_features = Channel.of([ exp_meta, file(params.features, checkIfExists: true)])
     } else if (params.study_type == 'affy_array'){
         ch_features = ch_affy_platform_features
+    } else if (params.study_type == 'maxquant'){
+
+        // For maxquant, we will use the processed matrices from PROTEUS
+        ch_features = ch_in_norm
```
pinin4fjords, line 227: This is very similar to what happens in the final part of this conditional, so let's combine the two rather than duplicating what happens there. Right now the last bit is

```
        // Otherwise we can just use the matrix input
        matrix_as_anno_filename = "matrix_as_anno.${matrix_file.getExtension()}"
        matrix_file.copyTo(matrix_as_anno_filename)
        ch_features = Channel.of([ exp_meta, file(matrix_as_anno_filename)])
```

But you could do something like:

```
        // Otherwise we can just use the matrix input

        matrix_as_anno_filename = "matrix_as_anno.${matrix_file.getExtension()}"
        if (params.study_type == 'maxquant'){
            ch_features_matrix = ch_in_norm
        } else{
            ch_features_matrix = ch_in_raw
        }
        ch_features = ch_features_matrix
            .map{ exp_meta, matrix_file -> 
                matrix_file.copyTo(matrix_as_anno_filename)
                return [exp_meta, file(matrix_as_anno_filename)]
            }
```

workflows/differentialabundance.nf
```diff
@@ -157,8 +177,35 @@ workflow DIFFERENTIALABUNDANCE {
...
+        // Split contrasts for proteus and for the later modules
+        ch_contrasts_split = ch_contrasts_file
+            .splitCsv ( header:true, sep:(params.contrasts.endsWith('tsv') ? '\t' : ','))
+            .map{ it.tail().first() }
+
+        // For proteus, extract only contrast variable as the module has to run once per contrast
+        ch_contrasts_proteus = ch_contrasts_split
+            .map{
+                tuple('id': it.variable)
+            }
```
pinin4fjords, line 191:
```suggestion

        // We'll be running Proteus once per unique contrast variable to generate plots
        // TODO: there should probably be a separate plotting module in proteus to simplify this
        
        ch_contrast_variables = ch_contrasts_file
            .splitCsv ( header:true, sep:(params.contrasts.endsWith('tsv') ? '\t' : ','))
            .map{ it.tail().first() }
            .map{
                tuple('id': it.variable)
            }
```

workflows/differentialabundance.nf
```diff
@@ -157,8 +177,35 @@ workflow DIFFERENTIALABUNDANCE {
...
+            .unique()
+
+        // Run proteus to import protein abundances
+        PROTEUS(
+            ch_contrasts_proteus.combine(proteus_in)
```
pinin4fjords, line 196:
```suggestion
            ch_contrast_variables.combine(proteus_in)
```

workflows/differentialabundance.nf
```diff
@@ -157,8 +177,35 @@ workflow DIFFERENTIALABUNDANCE {
...
+        )
+
+        // Re-map the proteus output tables to the study ID as the tables are the same across contrasts
+        ch_in_raw = PROTEUS.out.raw_tab
+            .reduce{a, b -> a}
+            .map{tuple('id': exp_meta.id, it[1])}
+        ch_in_norm = PROTEUS.out.norm_tab
+            .reduce{a, b -> a}
+            .map{tuple('id': exp_meta.id, it[1])}
```
pinin4fjords, line 205:
```suggestion
        ch_in_raw = PROTEUS.out.raw_tab
            .first()
            .map{ meta, matrix -> tuple(exp_meta, matrix) }
            
        ch_in_norm = PROTEUS.out.norm_tab
            .first()
            .map{ meta, matrix -> tuple(exp_meta, matrix) }
```

I think this is much simpler to understand.

pinin4fjords (changes requested):
Getting there- just a few more simplifications to facilitate future maintenance of these changes.

workflows/differentialabundance.nf
WackerO: TIL that the inside of the map is not separate from the rest of the code. I had to rename `exp_meta, matrix_file` and  `return [exp_meta` to not receive this error:
`- cause: The current scope already contains a variable of the name exp_meta`

workflows/differentialabundance.nf
WackerO: Oop, I did not realize `first()` has the same effect as my `reduce()` snippet. Thanks!

workflows/differentialabundance.nf
pinin4fjords: yep, my bad

assets/differentialabundance_report.Rmd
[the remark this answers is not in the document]
WackerO: Ah of course, just had to update shinyngs for the rmarkdown report for that function to be found. I have for now added the [] removal in the report; maybe in a future release that part can be integrated into shiny as well

pinin4fjords (approved):
OK, think we're good, much better integrated now, thanks for the work.

WackerO:
Thanks for the review!

PR #193, opened 2023-11-08
Add DESeq text to report

WackerO:
This PR adds a bit of explanatory text to the report about the DE analysis; for now only if DESeq2 is run:
[SRP254919_text.html.zip](https://github.com/nf-core/differentialabundance/files/13297446/SRP254919_text.html.zip)

## PR checklist

### DIFF
--- a/assets/differentialabundance_report.Rmd
+++ b/assets/differentialabundance_report.Rmd
@@ -882,22 +882,6 @@ make_params_table('exploratory analysis', 'exploratory_', remove_pattern = TRUE)
 
 ## Differential analysis
 
-```{r, echo=FALSE, results='asis', eval=params$study_type %in% c('rnaseq')}
-# For DESeq2, add some more explanation to the report
-cat(paste0(
-"The `DESeq2 R` package was used for differential analysis. p-values were adjusted with the ",
-params$deseq2_p_adjust_method,
-" method to reduce the number of false positives (",
-params$features_type,
-"s that are not actually differential). ",
-ucfirst(params$features_type),
-"s were considered differential if, for the respective contrast, the adjusted p-value was equal to or lower than ",
-params$deseq2_alpha,
-" and the absolute log2 fold change was equal to or higher than ",
-params$deseq2_lfc_threshold,
-"."
-))
-```
 
 ```{r, echo=FALSE, results='asis'}
 if (params$study_type == 'rnaseq'){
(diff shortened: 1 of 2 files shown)


assets/differentialabundance_report.Rmd
```diff
@@ -671,11 +671,36 @@ foo <- lapply(informative_variables[iv_min_group_sizes > 2], function(iv){
...
+```
+```{r, echo=FALSE, results='asis'}
+cat(paste0(
+"The following tables and volcano plots show how many and which ",
+params$features_type,
+"s were found to be differential in each contrast. In the plots, a ",
+params$features_type,
+" is colored differently depending on whether it was determined not differential, or differential according to log2 foldchange, or differential according to adjusted p-value, or differential according to both."
+))
+```
```
pinin4fjords, line 702:
```suggestion
```
Sorry, but I think this should be obvious from plot and table labels - don't want to add to much extraneous text.

assets/differentialabundance_report.Rmd
```diff
@@ -671,11 +671,36 @@ foo <- lapply(informative_variables[iv_min_group_sizes > 2], function(iv){
...
+" method to reduce the number of false positives (",
+params$features_type,
+"s that are not actually differential). ",
+ucfirst(params$features_type),
+"s were considered differential if, for the respective contrast, the adjusted p-value was equal to or lower than ",
+params$deseq2_alpha,
+" and the absolute log2 fold change was equal to or higher than ",
+params$deseq2_lfc_threshold,
+"."
+))
```
pinin4fjords, line 692:
```suggestion
# For DESeq2, add some more explanation to the report
cat(paste0(
"The `DESeq2 R` package was used for differential analysis. p-values were adjusted with the ", params$deseq2_p_adjust_method, " method to reduce the number of false positives. ", ucfirst(params$features_type), "s were considered differential if, for the respective contrast, the adjusted p-value was equal to or lower than ", params$deseq2_alpha, " and the absolute log2 fold change was equal to or higher than ", params$deseq2_lfc_threshold, "."
))
```

I think we should expect people to understand what a false positive is.

assets/differentialabundance_report.Rmd
```diff
@@ -857,6 +882,23 @@ make_params_table('exploratory analysis', 'exploratory_', remove_pattern = TRUE)
...
+```
```
pinin4fjords, line 900:
```suggestion
```
If you repeat the content in two places then we have to maintain it in two places- and I don't think the repetition is useful anyway, so let's not.

assets/differentialabundance_report.Rmd
WackerO: I agree! I just took the text that was written by someone else from another pipeline :'D

pinin4fjords (approved):

PR #199, opened 2023-11-16
Add gprofiler

WackerO:
## PR checklist

- [x] `README.md` is updated (including new tool citations and authors/contributors).


workflows/differentialabundance.nf
```diff
@@ -563,15 +615,23 @@ workflow DIFFERENTIALABUNDANCE {
         params.exploratory_assay_names.split(',').collect { "${it}_matrix".toString() } +
         [ 'contrasts_file', 'versions_file', 'logo', 'css', 'citations' ]
 
+
     // Condition params reported on study type
 
-    def params_pattern = ~/^(report|study|observations|features|filtering|exploratory|differential|deseq2|gsea).*/
+    def params_pattern = "report|study|observations|features|filtering|exploratory|differential|gsea|round_digits"
```
pinin4fjords, line 621: I think you should prefix the round digits param to place it into into one of the existing categories (maybe the report?)

workflows/differentialabundance.nf
```diff
@@ -563,15 +615,23 @@ workflow DIFFERENTIALABUNDANCE {
...
+    if (params.study_type == 'rnaseq'){
+        params_pattern += "|deseq2"
+    }
     if (params.study_type == 'affy_array' || params.study_type == 'geo_soft_file'){
-        params_pattern = ~/^(report|study|observations|features|filtering|exploratory|differential|affy|limma|gsea).*/
+        params_pattern += "|affy|limma"
```
pinin4fjords, line 626: nice- tidy

pinin4fjords (changes requested):
I'm going to have a another run-through of this in the morning. But for now: do you think it's sensible to have BOTH GSEA and gprofiler run? We could have a variety of gene set methods available, and I'm inclined to have one active at a time or the report will get very busy and complex.

WackerO:
> I'm going to have a another run-through of this in the morning. But for now: do you think it's sensible to have BOTH GSEA and gprofiler run? We could have a variety of gene set methods available, and I'm inclined to have one active at a time or the report will get very busy and complex.

Good point, but I would prefer not to outright prevent multiple methods from running simultaneously (sb might want to compare the results for example). I feel like if all of these methods are disabled by default, that should be fine, no? If a user then decides to run multiple ones in parallel, that's sort of on them :'D

workflows/differentialabundance.nf
WackerO: Ah yes, will do. If we add more of these params that affect multiple modules, we might consider adding a new category

assets/differentialabundance_report.Rmd
```diff
@@ -871,6 +914,42 @@ if (any(unlist(params[paste0(possible_gene_set_methods, '_run')]))){
 }
 ```
 
+```{r, echo=FALSE, results='asis', eval=params$gprofiler2_run}
+enrichment_files <- grep("gprofiler2", list.files(params$input_dir), value=T, fixed=T)
+cat(paste0("\n### Pathway enrichment analysis {.tabset}"))
```
pinin4fjords, line 919: This needs to be within the contrast-wise loop above- you'll see from the structure there that I anticipated multiple gene set methods in future. You should be able to infer the files you'll need to read from the contrast variables available there, rather than inferring contrasts from the gprofiler files.

docs/usage.md
```diff
@@ -269,6 +269,18 @@ With this configuration in place deployment should happen automatically every ti
 
 There is also a [Shiny server application](https://posit.co/download/shiny-server/), which you can install on your own infrastruture and use to host applications yourself.
 
+## Pathway enrichment analysis
```
pinin4fjords, line 272: Could you include GSEA here? At least mention it as an option/ add a subheading so that I can go back and fill it in (I didn't realise I hadn't documented that at all).

nextflow_schema.json
```diff
@@ -903,6 +991,20 @@
             },
             "fa_icon": "fab fa-app-store-ios"
         },
+        "multi_module_settings": {
+            "title": "Multi module settings",
+            "type": "object",
+            "description": "These parameters affect multiple modules in the pipeline.",
+            "default": "",
+            "properties": {
+                "round_digits": {
```
pinin4fjords: Just use the existing 'reporting_options' section and mention in the description that it just affects some tabular outputs in the report HTML.

workflows/differentialabundance.nf
```diff
@@ -117,6 +128,7 @@ include { LIMMA_DIFFERENTIAL                                } from '../modules/n
 include { CUSTOM_MATRIXFILTER                               } from '../modules/nf-core/custom/matrixfilter/main'
 include { ATLASGENEANNOTATIONMANIPULATION_GTF2FEATUREANNOTATION as GTF_TO_TABLE } from '../modules/nf-core/atlasgeneannotationmanipulation/gtf2featureannotation/main'
 include { GSEA_GSEA                                         } from '../modules/nf-core/gsea/gsea/main'
+include { GPROFILER2_GOST as GOST                           } from '../modules/nf-core/gprofiler2/gost/main'
```
pinin4fjords, line 131: could we leave it as "GPROFILERS_GOST" for consistency with "GSEA_GSEA"?

nextflow_schema.json
```diff
@@ -864,6 +859,99 @@
...
+            "type": "object",
+            "description": "",
+            "default": "",
+            "properties": {
+                "gprofiler2_run": {
+                    "type": "boolean",
+                    "description": "Set to run gprofiler2 and do a pathway enrichment analysis.",
+                    "fa_icon": "fas fa-running"
+                },
+                "gprofiler2_organism": {
```
pinin4fjords, line 874: OK for now, but when we sort out references I think we may need an optional generic species label at the top level of the report.

workflows/differentialabundance.nf
```diff
@@ -454,6 +466,38 @@ workflow DIFFERENTIALABUNDANCE {
...
+            ch_background = []
+        } else if (params.gprofiler2_background_file == "auto") {
+            // If auto, use input matrix as background
+            if(params.study_type == "geo_soft_file") {
+                ch_background = ch_norm.map{it.tail()}
+            } else {
+                ch_background = ch_raw.map{it.tail()}
+            }
+        } else {
+            if (!file(params.gprofiler2_background_file).exists()) {
```
pinin4fjords, line 482: You can just use `file(param, checkIfExists: true)`

workflows/differentialabundance.nf
```diff
@@ -454,6 +466,38 @@ workflow DIFFERENTIALABUNDANCE {
...
+                error("Param gprofiler2_background_file must be null, 'auto' or valid file path!")
+            }
+            ch_background = Channel.fromPath(params.gprofiler2_background_file)
+        }
+        if (!params.gprofiler2_gmt_file) {
```
pinin4fjords, line 487: There is already a param `gsea_gene_sets` for providing GMT files to the pipeline. Let's just make that more generic so you can use it (and the `ch_gene_sets` channel) for gprofiler too.

workflows/differentialabundance.nf
```diff
@@ -531,6 +575,14 @@ workflow DIFFERENTIALABUNDANCE {
             )
     }
 
+    if (params.gprofiler2_run){
+        ch_report_input_files = ch_report_input_files
+            .combine(GOST.out.plot_html.map{it[1]}.flatMap().toList())
+            .combine(GOST.out.all_enrich.map{it[1]}.flatMap().toList())
+            .combine(GOST.out.sub_enrich.map{it[1]}.flatMap().toList())
+        GOST.out.plot_html
```
pinin4fjords, line 583: Should there be a combine here?

pinin4fjords (commented):
Some more comments.

Mostly I think gprofiler should be better integrated into existing structures for working with gene sets (gene set file channels, report sections etc).

assets/differentialabundance_report.Rmd
WackerO: @pinin4fjords I'm wondering if we want to differentiate between gene set enrichment analysis and pathway enrichment analysis...if so, we should keep separate paragraphs in the report; otherwise I can add gprofiler into the existing loop

nextflow_schema.json
WackerO: Yep; ATM only gprofiler uses that param, that's why I made it specific, but can this change in the future

workflows/differentialabundance.nf
WackerO: Oops, that was a copy-paste-error, the channel is already combined in that block

docs/usage.md
WackerO: Will do!

assets/differentialabundance_report.Rmd
WackerO: I think pathway is more function-focussed, gene set sounds broader and not as specific to a certain purpose. I also looked the terms up, seems they are indeed (at least sometimes) distinguished, like [here](https://advaitabio.com/science/pathway-analysis-vs-gene-set-analysis/) and [here](https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-015-0751-5) (abstract).
But I don't specifically mind either way. However, I just realized that the GMT files may look different for GSEA and gprofiler2, so if we move the two together into one category and e.g. use only a single GMT parameter, I think we might have to keep those separate...

pinin4fjords: So, functionally these things are the same thing. We're talking sets of genes whether you mean pathways, GO terms, whatever- so let's keep to gene set (unless you propose limiting users to 'pathway' gene sets in some way). 

Could you expand on the GMT file differences? I would expect them to be interchangeable, so if they are indeed different we should document what the blocking differences are for each method- e.g. why a gene sets file I used for GSEA couldn't also be used for gprofiler.

pinin4fjords: Fundamentally the database shouldn't matter, these are just 'bags of genes', with identifier types corresponding to the ones you use in your input matrices. The source is irrelevant at run time (unless gprofiler is doing something I don't understand).

pinin4fjords: And no, there's no gene set database invoked at runtime in GSEA as we're using it- we supply the gene sets via the GMT files and GSEA does the stats.

WackerO:
Note: This can only be merged after https://github.com/nf-core/modules/pull/4538

assets/differentialabundance_report.Rmd
WackerO: I added gprofiler to the GSEA section. I also created a new param group called mods_ into which we can put all parameters that affect multiple modules (I'm open for suggestions for better names) and added both round_digits and gene_sets into that category; I felt like gene_sets does not really fit report_. There is also a short documentation section about GSEA; if you want, you can also send me some more elaborate text and I'll add that in

assets/differentialabundance_report.Rmd
```diff
@@ -838,32 +880,67 @@ for (i in 1:nrow(contrasts)){
...
+            cat("\n###### ", contrast_descriptions[i], "\n")
+            target_gsea_results <- read_metadata(target_gsea_tables[i])[,c(-2,-3)]
+            print( htmltools::tagList(datatable(target_gsea_results, caption = paste0("\nTarget (", contrasts$target[i], ")\n"), rownames = FALSE) ))
+            ref_gsea_results <- read_metadata(reference_gsea_tables[i])[,c(-2,-3)]
+            print( htmltools::tagList(datatable(ref_gsea_results, caption = paste0("\nReference (", contrasts$reference[i], ")\n"), rownames = FALSE) ))
+          }
+        }
+
+      } else if (gene_set_method == 'gprofiler2') {
+        enrichment_files <- grep("gprofiler2", list.files(params$input_dir), value=T, fixed=T)
```
pinin4fjords, line 913: Can you be consistent with the GSEA loop please? So loop through the gmt files and contrasts in the same way, and infer what the gprofiler files should be from the contrast names (they should be predictable). Then you can say explicitly when things are missing for a given contrast or gene sets file.

assets/differentialabundance_report.Rmd
```diff
@@ -838,32 +880,67 @@ for (i in 1:nrow(contrasts)){
...
+
+          for (html in rev(grep(".html", enrichment_files, value=T, fixed=T))) {
+            contrast <- unlist(strsplit(html, "gprofiler2.", fixed=T))[2]
+            contrast <- unlist(strsplit(contrast, ".gostplot", fixed=T))[1]
+            if (! file.exists(file.path(params$input_dir, html))){
+              stop(paste("gprofiler2 gost plot", file.path(params$input_dir, html), "does not exist"))
+            }
+
+            cat(paste0("\n##### ", contrast, "\n"))
+            cat(paste0('<embed type="text/html" src="', file.path(params$input_dir, html), '" width="800" height="400">'))
```
pinin4fjords, line 926: I'm not convinced by embedding random HTML in here, it could get messy and we're putting a lot of trust in the people who make those HTML files not to make them huge and unweildy. 

I could have done that for GSEA, but intentionally kept things to the summary tables and I think we should do the same for this. The user can go and refer to the individual HTML reports themselves.

assets/differentialabundance_report.Rmd
```diff
@@ -937,6 +1014,12 @@ if (any(unlist(params[paste0(possible_gene_set_methods, '_run')]))){
 }
 ```
 
+```{r, echo=FALSE, results='asis', eval=params$gprofiler2_run}
```
pinin4fjords, line 1017: This versions table should happen automatically as part of the above - you shouldn't need a new block

pinin4fjords (changes requested):
Getting closer, but as per the comments I think the integration and consistency with what is done for GSEA could go further. 

Also, for the subway map it's not the case that workflow would flow from one gene set method to the next. Could you split them out horizontally so the path branches to the two different methods and rejoins after?

nextflow.config
```diff
@@ -168,6 +181,10 @@ params {
     shinyngs_shinyapps_account      = null
     shinyngs_shinyapps_app_name     = null
 
+    // Multi-module options
+    mods_gene_sets                  = null
```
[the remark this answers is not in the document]
WackerO: Actually, to clarify: So the section will be called gene_set_files? Or the parameter? If the latter, in which section should it go?

assets/differentialabundance_report.Rmd
WackerO: The GMT file is optional for gprofiler; it does not behave exactly like the GSEA module. I'll have to think how to still make the loop work

pinin4fjords: So gprofiler has internal databases it would use without gmt files? Fair enough then- but still iterate over contrasts and look for matching files I think.

WackerO:
I believe I implemented all requested changes. Hope this is fine now; not sure how similar I can make the code for gsea and gprofiler in the report given that the two modules function a bit different

conf/test.config
```diff
@@ -45,7 +45,7 @@ params {
     // Test dataset is too small for the nsub default value
     deseq2_vst_nsub                   = 500
 
-    // Activate GSEA
-    gsea_run = true
-    gsea_gene_sets = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/gene_set_analysis/mh.all.v2022.1.Mm.symbols.gmt'
+    // Activate gprofiler2
```
pinin4fjords, line 48: Could you not activate this in the small test profile please? I left GSEA out of this to keep it minimal. Fine for test_full.

Also, when I tried to run GSEA at the same time by activating:

```
   // Activate GSEA
    gsea_run = true
    gene_sets_files = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/gene_set_analysis/mh.all.v2022.1.Mm.symbols.gmt'
```

I got this error:

```
  Your custom annotations ID is gp__NSxz_Dr8C_16U
  You can use this ID as an 'organism' name in all the related enrichment tests against this custom source.
  Just use: gost(my_genes, organism = 'gp__NSxz_Dr8C_16U')
  Detected custom GMT source request
  No results to show
  Please make sure that the organism is correct or set significant = FALSE
```

pinin4fjords (changes requested):
Thanks for the work. A few more things:

- I'd like to avoid doing list.files() test at the top- it's better if the user sees a heading for each gene set test and a message where the results file is empty. That's consistent with the "No significantly differential" message in the differential gene tables. When I get time I'll check what the GSEA is doing for empty results (it's probably just showing empty tables, and a message would be better).
- Best not to activate this in the test profile and keep that minimal 
- There seems to be an error in gProfiler when you activate GSEA and provide the gene sets - possibly some missing logic. Could you address that? 
- Could you make sure that `test_full` runs please? The gene sets param hasn't been updated there, so I'm assuming you haven't yet. It's also important that we demonstrate that gProfiler can run alongside GSEA.

docs/usage.md
```diff
@@ -269,6 +269,30 @@ With this configuration in place deployment should happen automatically every ti
 
 There is also a [Shiny server application](https://posit.co/download/shiny-server/), which you can install on your own infrastruture and use to host applications yourself.
 
+## Gene set enrichment analysis
+
+Currently, two tools can be used to do gene set enrichment analysis.
+
+### g:Profiler
```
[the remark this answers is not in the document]
WackerO: Nope, the tool itself is called [g:Profiler](https://academic.oup.com/nar/article/47/W1/W191/5486750); I decided to add the name of the tool itself instead of the R package gprofiler2. Should I change that?

I went for an alphabetic ordering, but sure, will change!

assets/differentialabundance_report.Rmd
```diff
@@ -838,32 +880,70 @@ for (i in 1:nrow(contrasts)){
...
+            cat("\n###### ", contrast_descriptions[i], "\n")
+            target_gsea_results <- read_metadata(target_gsea_tables[i])[,c(-2,-3)]
+            print( htmltools::tagList(datatable(target_gsea_results, caption = paste0("\nTarget (", contrasts$target[i], ")\n"), rownames = FALSE) ))
+            ref_gsea_results <- read_metadata(reference_gsea_tables[i])[,c(-2,-3)]
+            print( htmltools::tagList(datatable(ref_gsea_results, caption = paste0("\nReference (", contrasts$reference[i], ")\n"), rownames = FALSE) ))
+          }
+        }
+
+      } else if (gene_set_method == 'gprofiler2') {
+        enrichment_files <- grep("gprofiler2", list.files(params$input_dir), value=T, fixed=T)
```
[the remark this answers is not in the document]
WackerO: aah yes sure, on it!

workflows/differentialabundance.nf
```diff
@@ -396,6 +406,16 @@ workflow DIFFERENTIALABUNDANCE {
             .map{ it.tail() }
     }
 
+    // We'll use a local module to filter the differential tables and create output files that contain only differential features
+    ch_logfc = Channel.value([ (params.study_type == 'rnaseq' ? 'log2FoldChange' : 'logFC'), params.differential_min_fold_change ])
```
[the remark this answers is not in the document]
WackerO: Ou you're right, that was an oversight!

workflows/differentialabundance.nf
```diff
@@ -454,6 +474,37 @@ workflow DIFFERENTIALABUNDANCE {
...
 
+    if (params.gprofiler2_run) {
+
+        // For gprofiler2, use only features that are considered differential
+        ch_filtered_diff = FILTER_DIFFTABLE.out.filtered
+
+        if (!params.gprofiler2_background_file) {
+            // If param not set, use empty list as "background"
+            ch_background = []
+        } else if (params.gprofiler2_background_file == "auto") {
```
pinin4fjords, line 485: I've been thinking about this, and I think that unless you have good reason for doing this, the background file should be the universe of genes considered for differential expression, which is the output of `CUSTOM_MATRIXFILTER`.

conf/test.config
WackerO: Hmm, sorry Jonathan, I can't reproduce that error...Did you just run the test.config with the `gsea_run` and `gene_sets_files` parameters added?

workflows/differentialabundance.nf
WackerO: Aah sure, you mean invert the default so that the user has to expressly deactivate the background? I think that makes sense!

pinin4fjords: Yes- I really thing the background is the right way to go, not doing that is a really common mistake and one we should very strongly discourage. 

The universe of expressed genes is the correct background to use (which we effectively derive in matrix filtering), so I think using CUSTOM_MATRIXFILTER makes sense.

conf/test.config
pinin4fjords: Yep!

WackerO: I added both gsea and gpro to the full test (this time with a token for gpro from a previous run; let's see how long g:profiler saves these tokens). For me it worked; can you try again?

modules/local/filter_difftable.nf
```diff
@@ -0,0 +1,61 @@
...
+
+    head -n 1 $tsv > \${outfile}.tmp
+
+    # The following snippet performs the following checks on each row (add +0.0 to the numbers so that they are definitely treated as numerics):
+    # 1. Check that the current logFC/padj is not NA
+    # 2. Check that the current logFC is >= threshold (abs does not work, so use a workaround)
+    # 3. Check that the current padj is <= threshold
+    # If this is true, the row is written to the new file, otherwise not
+    tail -n +2 $tsv | awk -F'\\t' -v logFC_col=\$logFC_col -v padj_col=\$padj_col '
+    \$logFC_col != "NA" && \$padj_col != "NA" && (\$logFC_col+0.0 >= 0 ? \$logFC_col+0.0 >= ENVIRON["logFC_threshold"]+0.0 : -(\$logFC_col+0.0) >= ENVIRON["logFC_threshold"]+0.0) && \$padj_col+0.0 <= ENVIRON["padj_threshold"]+0.0 {print}' >> \${outfile}.tmp
```
pinin4fjords, line 52:
```suggestion
output_file="\${input_file%.*}_filtered.tsv"

# Function to find column number
find_column_number() {
    awk -v column="\$2" '{for(i=1;i<=NF;i++) if (\$i == column) {print i; exit}}' <<< "\$(head -n 1 "\$1")"
}

# Extract column numbers
logFC_col=\$(find_column_number "\$input_file" "log2FoldChange")
padj_col=\$(find_column_number "\$input_file" "padj")

# Prepare the output file
head -n 1 "\$input_file" > "\${output_file}.tmp"

# The following snippet performs the following checks on each row (add +0.0 to the numbers so that they are definitely treated as numerics):
#
# 1. Check that the current logFC/padj is not NA
# 2. Check that the current logFC is >= threshold (abs does not work, so use a workaround)
# 3. Check that the current padj is <= threshold
#
# If this is true, the row is written to the new file, otherwise not
    
awk -F'\\t' -v logFC_col="\$logFC_col" -v padj_col="\$padj_col" -v logFC_thresh="\$logFC_threshold" -v padj_thresh="\$padj_threshold" '
    NR > 1 && \$logFC_col != "NA" && \$padj_col != "NA" &&
    ((\$logFC_col+0.0 >= logFC_thresh+0.0) || (-\$logFC_col+0.0 >= logFC_thresh+0.0)) &&
    \$padj_col+0.0 <= padj_thresh+0.0 { print }
' "\$input_file" >> "\${output_file}.tmp"

# Rename temporary file to final output file
mv "\${output_file}.tmp" "\$output_file"
```
I think something like this is more readable, using the -v syntax for injecting variables into the awk

docs/usage.md
```diff
@@ -269,6 +269,30 @@ With this configuration in place deployment should happen automatically every ti
...
+[GSEA](https://www.gsea-msigdb.org/gsea/index.jsp) tests for differential genes from within a user-provided set of genes; this requires a GMT or GMX file. The following example shows how to enable this:
+
+```bash
+--gsea_run true \
+--gene_sets_files gene_sets.gmt
+```
+
+### g:Profiler
+
+The [gprofiler2](https://cran.r-project.org/web/packages/gprofiler2/vignettes/gprofiler2.html) package can be used to test which pathways are enriched for genes that have been determined to be differentially expressed; it is an R interface for the g:Profiler webtool. In the simplest form, this feature can be enabled with the parameters from the following example:
```
pinin4fjords, line 287: > test which pathways are enriched for genes that have been determined to be differentially expressed

Is this true? Or is it rather looking for which pathways are enriched in the set of differential genes relative to the background? Not quite the same thing.

docs/usage.md
```diff
@@ -269,6 +269,30 @@ With this configuration in place deployment should happen automatically every ti
...
+
+```bash
+--gprofiler2_run true \
+--gprofiler2_organism mmusculus
+```
+
```
pinin4fjords, line 293: Add a note on the background, how important that is, that 'auto' is the default and what that means?

modules/local/filter_difftable.nf
pinin4fjords: (I did check this worked in a task directory, hope I've got all the escapes right here)

pinin4fjords (changes requested):
Really close, just a few last comments and I tidied up the bash filtering script

assets/differentialabundance_report.Rmd
```diff
@@ -838,32 +880,60 @@ for (i in 1:nrow(contrasts)){
...
```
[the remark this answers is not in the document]
WackerO: Oh you're right!

conf/test_full.config
```diff
@@ -34,5 +34,9 @@ params {
 
     // Activate GSEA
     gsea_run = true
-    gsea_gene_sets = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/gene_set_analysis/mh.all.v2022.1.Mm.symbols.gmt'
+    gene_sets_files = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/gene_set_analysis/mh.all.v2022.1.Mm.symbols.gmt'
+
+    // Activate gprofiler2
+    gprofiler2_run = true
+    gprofiler2_token = 'gp__0YRX_1WHk_VjE'
```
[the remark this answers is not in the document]
WackerO: Unfortunately I haven't found any info on how long exactly a token is available, but it should definitely last for some time (after it expires, I can generate a new one; not ideal of course).
I can also leave it out, then the gene sets will be used, but gprofiler does not find results with the gene sets file and I thought it would be good to have a test config in which gprofiler generates some pathway results.

modules/local/filter_difftable.nf
WackerO: I got an unbound variable error so I had to change some escapes.
However, with the output_file definition from your code, I could not figure out a way to make the escape work within the nextflow module; either it would throw an error or it would just not correctly do the string replacement, so I had to go back to the previous solution with sed (it's the same as in the tabular_to_gsea module).

docs/usage.md
WackerO: Umm, I'm not sure I understand. I was meaning that gprofiler searches for enrichment of pathways with DE genes as it only receives those (it is given the filtered matrix, not the full results from DESeq2/limma).

docs/usage.md
WackerO: Ah good point!

docs/usage.md
pinin4fjords: So, your wording suggests that you start with a gene set and try to understand whether there are more differentially expressed genes that you would expect by chance. I suspect that's not what's happening.

Rather, we take the differentially expressed gene list, and look to see if there are more members of each gene set than we would expect by chance, relative to the background. The background contains all the genes passed to d/e, so it effectively does have the full results from d/e (though all it knows is that they are not significant).

conf/test_full.config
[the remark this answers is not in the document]
pinin4fjords: OK, we can't have an ephemeral token (it will almost certainly have finite life) that can just expire and invalidate tests, that's just madness.

Why wouldn't gprofiler find results with just the gene sets? What's the functional difference in what is done with the token?

modules/local/filter_difftable.nf
pinin4fjords: Looks good- glad the rest of it worked

docs/usage.md
WackerO: Aah, okay, maybe I formulated that badly. I wanted to indicate that only those genes are piped into the enrichment analysis that DESeq/limma deemed differential. Maybe something like this instead?

`The [gprofiler2](https://cran.r-project.org/web/packages/gprofiler2/vignettes/gprofiler2.html) package can be used to test which pathways are enriched for genes that the DESeq2/limma modules determined to be differentially expressed; it is an R interface for the g:Profiler webtool.`

pinin4fjords: That has the same ambiguity to me. How about something like:

The [gprofiler2](https://cran.r-project.org/web/packages/gprofiler2/vignettes/gprofiler2.html) package can be used to test which pathways are enriched in the set of differential genes produced by the the DESeq2 or limma modules.  It is an R interface for the g:Profiler webtool.

conf/test_full.config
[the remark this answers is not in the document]
WackerO: I am really not sure about this, but I suspect that gprofiler is only familiar with those datasources that the server offers (REAC, KEGG, GO etc.); the GMT in the test contains a bunch of hallmark entries and it is possible that gprofiler does not know what to do with those (for that GMT, I also cannot get any results with the [online service](https://biit.cs.ut.ee/gprofiler/gost)).

The token I added is simply from a previous gprofiler run which queried its normal databases, that's why it will find results there. But I'll remove the token. For now, the test will then only produce empty tables (which can of course also happen for an actual analysis).
I can sometime add some additional entries to the GMT in the config so that it will work for both gsea and gprofiler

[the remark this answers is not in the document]
pinin4fjords: I'm not totally fussed with it using the same gene sets as GSEA. 

How would you run this module in such a way that it would 'query its normal databases' without using the token?

docs/usage.md
WackerO: That sounds good, thanks!

conf/test_full.config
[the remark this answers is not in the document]
WackerO: I'd give it a GMT that contains some of the sources listed in gprofiler like GO or REAC. Again, it's possible that gprofiler for some reason just happens to not find any enrichment in the current GMT but would be totally fine with some other HALLMARK pathways, I'm not sure and wasn't able to find out for certain why it does not find enrichment; that is just my guess...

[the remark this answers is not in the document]
pinin4fjords: But you can run gprofiler here **without** supplying the gene sets, right? Weren't you doing that before? And that produced some results? And then it would use just its in-built gene sets, as you did in the UI? 

If I've understood that correctly, you might need to add an additional 'gprofiler_use_gene_sets' flag or somesuch such that it ignores any globally defined gene sets when specifified.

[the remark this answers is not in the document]
WackerO: Done!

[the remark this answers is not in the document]
pinin4fjords: I'm not seeing it- could you clarify?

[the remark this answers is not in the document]
WackerO: That'll be because I forgot to push -.- Sorry, this time it is _actually_ done

workflows/differentialabundance.nf
```diff
@@ -454,6 +482,32 @@ workflow DIFFERENTIALABUNDANCE {
...
+            // If deactivated, use empty list as "background"
+            ch_background = []
+        } else if (params.gprofiler2_background_file == "auto") {
+            // If auto, use input matrix as background
+            ch_background = CUSTOM_MATRIXFILTER.out.filtered.map{it.tail()}.first()
+        } else {
+            ch_background = Channel.from(file(params.gprofiler2_background_file, checkIfExists: true))
+        }
+
+        ch_gprofiler_gene_sets = ch_gene_sets
```
pinin4fjords, line 500: I fixed this so as not to overwrite the global var. As mentioned above, this hack would not be necessary if we arranged it so that the module prioritises the organism and token options. We would just supply the gene sets channel(empty or otherwise) and allow the module to ignore them in favour of other options.

workflows/differentialabundance.nf
```diff
@@ -66,12 +69,25 @@ if (params.study_type == 'affy_array'){
 // Check optional parameters
 if (params.transcript_length_matrix) { ch_transcript_lengths = Channel.of([ exp_meta, file(params.transcript_length_matrix, checkIfExists: true)]).first() } else { ch_transcript_lengths = [[],[]] }
 if (params.control_features) { ch_control_features = Channel.of([ exp_meta, file(params.control_features, checkIfExists: true)]).first() } else { ch_control_features = [[],[]] }
-if (params.gsea_run) {
-    if (params.gsea_gene_sets){
-        gene_sets_files = params.gsea_gene_sets.split(",")
+
+def run_gene_set_analysis = params.gsea_run || params.gprofiler2_run
+
+if (run_gene_set_analysis) {
```
pinin4fjords, line 75: I combined and simplified the gene set logic.

pinin4fjords (changes requested):
I added a commit - hope you don't mind, just check you agree.

At a minimum we should document the new option in usage.

Alternatively, and just a suggestion, we reorder the priorities in the gprofiler2 module, so that the new mode option isn't necessary (because the gprofiler2 module would simply ignore the gene sets in the presence of one of the other options - we'd just have to document that). 

What do you think?

workflows/differentialabundance.nf
WackerO: Ah, good catch with the global!

docs/usage.md
```diff
@@ -269,6 +269,30 @@ With this configuration in place deployment should happen automatically every ti
...
+### g:Profiler
+
+The [gprofiler2](https://cran.r-project.org/web/packages/gprofiler2/vignettes/gprofiler2.html) package can be used to test which pathways are enriched in the sets of differential genes produced by the the DESeq2 or limma modules. It is an R interface for the g:Profiler webtool. In the simplest form, this feature can be enabled with the parameters from the following example:
+
+```bash
+--gprofiler2_run true \
+--gprofiler2_organism mmusculus
+```
+
+Either a token from a previous g:profiler run, a custom GMT file or the organism (mmusculus for Mus musculus, hsapiens for Homo sapiens etc.) is required. Check the [pipeline webpage](https://nf-co.re/differentialabundance/1.2.0/parameters#gprofiler2) for the parameters for this step. Also, by default the analysis will be run with a background list of genes that passed the abundance filter (i.e. those genes that actually had some expression); see for example https://doi.org/10.1186/s13059-015-0761-7 for why this is advisable. You can provide your own background list with `--gprofiler2_background_file background.txt`; alternatively, if you want to not use any background, set this: `--gprofiler2_background_file false`.
```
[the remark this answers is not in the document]
WackerO: So I did set the original priorities intentionally (first the specific/reproducible options token and custom GMT, then the organism which just queries whichever database versions are the current ones). But update the module to put the GMT file as last instance.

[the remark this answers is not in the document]
WackerO: Done. I had to do some fiddling to keep the error message if multiple GMT files are provided. I'll some time update the module again so that the pipeline can be run with multiple GMT files like GSEA

workflows/differentialabundance.nf
```diff
@@ -454,6 +478,35 @@ workflow DIFFERENTIALABUNDANCE {
...
+        } else {
+            ch_gprofiler_gene_sets = ch_gene_sets.toList()
+                .map{it -> it.size() > 1 ? error("gprofiler2 currently only accepts a single gene sets file!") : it.first()}            
+        }
+
+        GPROFILER2_GOST(
+            ch_filtered_diff,
+            ch_gprofiler_gene_sets,
+            ch_background
+        )
```
pinin4fjords, line 508:
```suggestion
        // For gprofiler2, token and organism have priority and will override a gene_sets file

        GPROFILER2_GOST(
            ch_filtered_diff,
            ch_gene_sets,
            ch_background
        )
```

I've often had trouble conditioning on channel sizes at runtime- I don't think it works (or at least not well). But we can do the check outside of the workflow logic along with the other input checks- see the other suggestion.

But the beauty of the priorities in the module is that you can just pass the gene sets straight through, and they'll be ignored if the other options are activated- so this gets a lot simpler.

workflows/differentialabundance.nf
```diff
@@ -66,12 +69,21 @@ if (params.study_type == 'affy_array'){
 // Check optional parameters
 if (params.transcript_length_matrix) { ch_transcript_lengths = Channel.of([ exp_meta, file(params.transcript_length_matrix, checkIfExists: true)]).first() } else { ch_transcript_lengths = [[],[]] }
 if (params.control_features) { ch_control_features = Channel.of([ exp_meta, file(params.control_features, checkIfExists: true)]).first() } else { ch_control_features = [[],[]] }
-if (params.gsea_run) {
-    if (params.gsea_gene_sets){
-        gene_sets_files = params.gsea_gene_sets.split(",")
+
+def run_gene_set_analysis = params.gsea_run || params.gprofiler2_run
+
+if (run_gene_set_analysis) {
```
pinin4fjords, line 75: If you really don't think you can handle running the gprofiler module multiple times for different gene set files, suggest modifying the validation logic like:

```
if (run_gene_set_analysis) {
    if (params.gene_sets_files) {
        gene_sets_files = params.gene_sets_files.split(",")
        ch_gene_sets = Channel.of(gene_sets_files).map { file(it, checkIfExists: true) }
        if (params.gprofiler2_run && (!params.gprofiler2_token && !params.gprofiler2_organism) && gene_sets_files.size() > 1) {
            error("gprofiler2 can only work with a single gene set file")
        }
    } else if (params.gsea_run) {
        error("GSEA activated but gene set file not specified!")
    } else if (params.gprofiler2_run) {
        if (!params.gprofiler2_token && !params.gprofiler2_organism) {
            error("To run gprofiler2, please provide a run token, GMT file or organism!")
        }
    } else {
        ch_gene_sets = []    // For methods that can run without gene sets
    }
}
```

(sorry for not making this a proper suggestion- GitHub wouldn't let me becuase of the deleted lines)

docs/usage.md
```diff
@@ -269,6 +269,30 @@ With this configuration in place deployment should happen automatically every ti
...
+
+```bash
+--gprofiler2_run true \
+--gprofiler2_organism mmusculus
+```
+
+Either a `--gprofiler2_token` from a previous g:profiler run, the `--gprofiler2_organism` (mmusculus for Mus musculus, hsapiens for Homo sapiens etc.) or a custom `--gene_sets_files` GMT file is required; if multiple options are provided, will prioritize token, then organism, then GMT file. Check the [pipeline webpage](https://nf-co.re/differentialabundance/1.2.0/parameters#gprofiler2) for the parameters for this step. Also, by default the analysis will be run with a background list of genes that passed the abundance filter (i.e. those genes that actually had some expression); see for example https://doi.org/10.1186/s13059-015-0761-7 for why this is advisable. You can provide your own background list with `--gprofiler2_background_file background.txt`; alternatively, if you want to not use any background, set this: `--gprofiler2_background_file false`.
```
pinin4fjords, line 294:
```suggestion
If gene sets have been specified to the workflow via `--gene_sets_files` these are used by default. Specifying  `--gprofiler2_organism` (mmusculus for Mus musculus, hsapiens for Homo sapiens etc.) will override those gene sets with g:profiler's own for the relevant species. `--gprofiler2_token` will override both options and use gene sets from a previous g:profiler run.  

By default the analysis will be run with a background list of genes that passed the abundance filter (i.e. those genes that actually had some expression); see for example https://doi.org/10.1186/s13059-015-0761-7 for why this is advisable. You can provide your own background list with `--gprofiler2_background_file background.txt`or if you want to not use any background, set `--gprofiler2_background_file false`.

Check the [pipeline webpage](https://nf-co.re/differentialabundance/parameters#gprofiler2) for a full listing of the relevant parameters. 
```

Simplifying suggestion, plus we shouldn't point to a specific version of the webpage (we'd forget to update it on release).

pinin4fjords (changes requested):
Last batch of corrections. With these addressed we're ready to go!

docs/usage.md
WackerO: Oh yes, good point!

workflows/differentialabundance.nf
WackerO: Ah yes, that is simpler of course.

No, I will certainly enable multiple GMTs in the future when I get around to it, just not right now (I'll mention that in the error).

pinin4fjords (approved):
Good to go- thanks for working with me on this, think things are much tighter now.

Issue #227, opened 2024-01-22
Plotting of data where the p-value is zero


kdivilov:
### Description of the bug

If the pipeline (tested only with rnaseq) infers differentially expressed genes with p-values that are equal to 0, they are not plotted in the volcano plot. While this makes sense mathematically (log(0)=undefined), they are the most significantly differentially expressed genes and so would be ones people would want to look at.

### System information

nf-core/differentialabundance version 1.4.0

WackerO:
Hmm, good catch! Fixed in #229, does that work?

PR #229, opened 2024-01-30
Report volcano plot will now also plot points with pval 0

WackerO:
In response to https://github.com/nf-core/differentialabundance/issues/227

## PR checklist

### DIFF
--- a/assets/differentialabundance_report.Rmd
+++ b/assets/differentialabundance_report.Rmd
@@ -851,18 +851,24 @@ for (i in 1:nrow(contrasts)){
     full_de$differential_status[de_pval] <- de_pval_label
     full_de$differential_status[de_fc & de_pval] <- de_pval_fc_label
     full_de$differential_status <- factor(full_de$differential_status, levels = c("Not significant", de_fc_label, de_pval_label, de_pval_fc_label), ordered = TRUE) # Factorize status so that non-significant is always first
-    # Define the thresholds we'll draw
 
+    # Prepare y data; include 0 p-values which will otherwise produce Inf because of log
+    full_de$y_data <- sapply(full_de[[pval_column]], function(y) {ifelse(y==0,0,-log10(y))})
+    inf_value <- ceiling(max(abs(full_de$y_data)) * 1.1)
+    full_de$y_data[full_de$y_data==0] <- inf_value
+
+    # Define the thresholds we'll draw
     hline_thresholds = vline_thresholds = list()
     hline_thresholds[[paste(pval_column, '=', p_value_thresholds[[pvt]])]] = -log10(p_value_thresholds[[pvt]])
+    hline_thresholds[[paste(pval_column, '= 0')]] = inf_value
     vline_thresholds[[paste(params$differential_fc_column, '<=', log2(params$differential_min_fold_change))]] = -log2(params$differential_min_fold_change)
     vline_thresholds[[paste(params$differential_fc_column, '>=', log2(params$differential_min_fold_change))]] = log2(params$differential_min_fold_change)
-    
+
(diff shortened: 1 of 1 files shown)


WackerO:
E.g. see this report: [SRP254919_zero.html.zip](https://github.com/nf-core/differentialabundance/files/14100240/SRP254919_zero.html.zip)

kdivilov:
I'm not sure if the example report was to only show how the new p-value=0 threshold on the volcano plot would look like or to show a complete working example. If the former, then it looks great. If the latter, there are a few issues:
1. RP23-4H14.1 is not in sig_differential for both p_value_types.
2. RP23-4H14.1 is not given de_pval_fc_label for the Adjusted p_value_types.
3. All genes with de_pval_fc_label (Jchain, Igkc, Ighg2b, Igha) except Pou2af1 have p values = 0, so their y-values on the volcano plot should be either ceiling(max(abs(-log10(0.0003))) * 1.1) or ceiling(max(abs(-log10(0.0489))) * 1.1) for the Unadjusted and Adjusted p_value_types, respectively in the "hND6 versus mCherry in treatment" volcano plot (similar issue with the other volcano plot).

After reviewing the code, I couldn't find the reason for these issues, but I'm also not familiar with this codebase.

WackerO:
It was mainly there for showing what the plot looks like, but the example is also working actually. RP23-4H14.1 is not DE according to the tables, why would you expect it to be marked as DE? I just had a look at the DE table, its entry is
```
gene_id	gene_name	gene_biotype	pvalue	padj	log2FoldChange	differential_status	y_data
ENSMUSG00000107358	RP23-4H14.1	TEC	0.05018516	0.9026969	-0.06082605	Not significant	0.0444580489717465
```

Also, sorry for the confusion, the tables below the plots are rounded. The entries where it says padj 0 aren't actually 0, but a very small number. I've attached the report again, this time just without rounding:

[SRP254919.html.zip](https://github.com/nf-core/differentialabundance/files/14122585/SRP254919.html.zip)

pinin4fjords:
I'm not 100% sold on this. The data are fundamentally outside the axes, and inventing a p value ceiling such that we can show such points as a smear at the extremes of the plot doesn't add much to my mind. 

The tables are there to see the gene lists. @WackerO we might want to refine the rounding logic you added to the reporting to use scientific notation rather than marking non-0 p values as 0.

WackerO:
> I'm not 100% sold on this. The data are fundamentally outside the axes, and inventing a p value ceiling such that we can show such points as a smear at the extremes of the plot doesn't add much to my mind.
> 
> The tables are there to see the gene lists. @WackerO we might want to refine the rounding logic you added to the reporting to use scientific notation rather than marking non-0 p values as 0.

I can definitely change the rounding so that the table is clearer, but I would like to argue in favor of the ceiling. This is something that is for example also done by [gprofiler](https://biit.cs.ut.ee/gprofiler/gost) (clicking on `random example`, then`run query` will produce a plot that's capped at 16).
I think the plots are there to visualize the data that are listed in the tables, and if some entries are frequently missing, IMO the visualization is lacking (especially if those missing points are the ones with the smallest p value).

Do you have a specific reason why you don't want these points added to the plot? @pinin4fjords

pinin4fjords:
Because such points can't be represented properly, they are by definition outside the plot boundaries. 

You can't just define an arbitrary ceiling- another point could come along with a lower p value than the ceiling. So you'd probably have to set a ceiling equal to the smallest non-zero p value in the data (or, say half the lowest p value). Then you end up distorting the plot for the points you _can_ represent. 

I'd much rather just show a warning that there are points with p values of 0 that can't be shown- and the user can go find those in the tables.

WackerO:
#232 solves this, closing this PR

PR #257, opened 2024-03-27
Some fixes before the upcoming release

WackerO:
added maxquant profile to nextflow.config, clarified some docu, made obs_name_col optional in report.rmd, ~~fixed FILTER_DIFFTABLE module (changed to python as AWK stopped filtering correctly)~~ (done in https://github.com/nf-core/differentialabundance/pull/264)

## PR checklist


WackerO:
Is it necessary to "unbump" the versions?

docs/usage.md
```diff
@@ -23,7 +23,11 @@ With the above in mind, running this workflow requires:
 --input '[path to samplesheet file]'
 ```
 
-This may well be the same sample sheet used to generate the input matrix. For example, in RNA-seq this might be the same sample sheet, perhaps derived from [fetchngs](https://github.com/nf-core/fetchngs), that was input to the [RNA-seq workflow](https://github.com/nf-core/rnaseq). It may be necessary to add columns that describe the groups you want to compare.
+This may well be the same sample sheet used to generate the input matrix. For example, in RNA-seq this might be the same sample sheet, perhaps derived from [fetchngs](https://github.com/nf-core/fetchngs), that was input to the [RNA-seq workflow](https://github.com/nf-core/rnaseq). It may be necessary to add columns that describe the groups you want to compare. The columns that the pipeline requires are:
+
+- a column listing the sample IDs (must be the same IDs as in the abundance matrix); in the example below it is called sample. For some study_types, this column might need to be filled in with file names, e.g. when doing an affymetrix analysis
```
pinin4fjords, line 28:
```suggestion
- a column listing the sample IDs (must be the same IDs as in the abundance matrix), in the example below it is called 'sample'. For some study_types, this column might need to be filled in with file names, e.g. when doing an affymetrix analysis.
```

docs/usage.md
```diff
@@ -23,7 +23,11 @@ With the above in mind, running this workflow requires:
...
+- one or more columns describing conditions for the differential analysis; in the example below it is called condition
```
pinin4fjords, line 29:
```suggestion
- one or more columns describing conditions for the differential analysis. In the example below it is called 'condition'
```

docs/usage.md
```diff
@@ -23,7 +23,11 @@ With the above in mind, running this workflow requires:
...
+- optionally one or more columns describing sample batches or similar which you want to be considered in the analysis; in the example below it is called batch.
```
pinin4fjords, line 30:
```suggestion
- optionally one or more columns describing sample batches or similar which you want to be considered in the analysis. In the example below it is called 'batch'
```

nextflow_schema.json
```diff
@@ -268,8 +268,9 @@
             "properties": {
                 "proteus_measurecol_prefix": {
                     "type": "string",
-                    "default": "LFQ intensity ",
-                    "description": "Prefix of the column names of the MaxQuant proteingroups table in which the intensity values are saved; the prefix has to be followed by the sample names that are also found in the samplesheet. Default: 'LFQ intensity '; take care to also consider trailing whitespace between prefix and samplenames."
+                    "default": "LFQ intensity_s",
```
pinin4fjords, line 271: This seems weird and brittle- too hacky for me. People are going to misunderstand this and get it wrong. We need a different solution - why not just make the proteus module allow for an optional space after the prefix?

nextflow_schema.json
```diff
@@ -1025,19 +1027,19 @@
                 },
                 "logo_file": {
                     "type": "string",
-                    "default": "docs/images/nf-core-differentialabundance_logo_light.png",
+                    "default": "${projectDir}/docs/images/nf-core-differentialabundance_logo_light.png",
```
pinin4fjords, line 1030: Looks fine in principle, but these have been working fine for ages- why the sudden need for `projectDir`?

pinin4fjords (changes requested):
Mostly OK, but sorry, I think the space thing is a hack, we need a better workaround.

nextflow_schema.json
WackerO: I generally prefer not having my code assume that a whitespace is necessary and instead just take directly what the user inputs...but I suppose here, the flexible whitespace addition is the best solution. I just updated the proteus module with those changes

nextflow_schema.json
WackerO: I have no idea either, but for some reason now, having the path without `projectDir` causes this error: https://github.com/nf-core/differentialabundance/actions/runs/8416574741/job/23043674976?pr=256

modules/local/filter_difftable.nf
```diff
@@ -2,10 +2,10 @@ process FILTER_DIFFTABLE {
 
     label 'process_single'
 
-    conda "conda-forge::gawk=5.1.0"
```
pinin4fjords, line 5: Re-writing this process in Python is a quite a major change to smuggle into a PR of small fixes ;-).

Can you say why the process was re-written, and demonstrate that the output is the same relative to previous?

workflows/differentialabundance.nf
```diff
@@ -400,7 +400,7 @@ workflow DIFFERENTIALABUNDANCE {
     }
 
     // We'll use a local module to filter the differential tables and create output files that contain only differential features
-    ch_logfc = Channel.value([ params.differential_fc_column, params.differential_min_fold_change ])
+    ch_logfc = Channel.value([ params.differential_fc_column, Math.log(params.differential_min_fold_change)/Math.log(2) ])
```
pinin4fjords, line 403: I'm not sure what you're doing here, but could you do it in the module?

pinin4fjords (changes requested):
I'm a bit nervous of a complete re-write of a module being smuggled into a set of minor pre-release fixes. I'm concerned it could break something.

Can you clarify why that rewrite is necessary here, and demonstrate clearly that it's doing the same thing?

modules/local/filter_difftable.nf
WackerO: If you prefer, I can also do the module change in a separate PR, that's not a problem!
Nope, the output changed; it was incorrect before, that's why I had to change it :') AWK suddenly filtered out all the genes and just produced an empty table (was not the case when I originally added the module).
I tried a lot to change it in AWK but it simply did not work and I have to be honest, I found the AWK code quite painful to deal with as I'm simply not experienced enough with it.

WackerO: Oh, but as a follow-up, I did check the output. The new module + [this code](https://github.com/nf-core/differentialabundance/pull/257/files#diff-93d05d71f7facfc0a15a9131995c3b783d0d6ac38ccfd0bf7b5e654410c5c7e3R403) makes it so that the results are the same as the DE genes according to the html report.

workflows/differentialabundance.nf
WackerO: Can do

modules/local/filter_difftable.nf
pinin4fjords: Actually, could you raise this as a separate PR? I think that would make more sense and you can post the reasoning and the evidence there.

WackerO: Yes of course, see #264

PR #264, opened 2024-04-23
Change FILTER_DIFFTABLE to python because AWK does not correctly filter reliably

WackerO:
This PR changes the FILTER_DIFFTABLE module to python with pandas; I noticed while working on #257 that the previous module sometimes filters out all the genes, not just the non-DE ones. I was not able to find where the issue is with AWK after a lot of trying out, so the module now uses python.
The module now also applies log2 to the threshold parameter so that the DE gene results are the same as in the report tables; I checked this and the tables now agree with each other (see attached zip).

[filtered_results.zip](https://github.com/nf-core/differentialabundance/files/15073896/filtered_results.zip)

## PR checklist

### DIFF
--- a/.nf-core.yml
+++ b/.nf-core.yml
@@ -1,1 +1,8 @@
 repository_type: pipeline
+lint:
+  nextflow_config:
+    - config_defaults:
+        - params.logo_file
+        - params.css_file
+        - params.citations_file
+        - params.report_file
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Changed`
 
+- [[#260](https://github.com/nf-core/differentialabundance/pull/260)] - Change FILTER_DIFFTABLE to python because AWK does not correctly filter reliablys ([@WackerO](https://github.com/WackerO), review by [@pinin4fjords](https://github.com/pinin4fjords))
 - [[#232](https://github.com/nf-core/differentialabundance/pull/232)] - Mention missing dots in volcano plot, change rounding, turn off rounding by default ([@WackerO](https://github.com/WackerO), review by [@pinin4fjords](https://github.com/pinin4fjords))
 
 ## v1.4.0 - 2023-11-27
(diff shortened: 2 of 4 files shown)


modules/local/filter_difftable.nf
```diff
@@ -20,42 +20,34 @@ process FILTER_DIFFTABLE {
...
+    table = table[~table["$logFC_column"].isna() &
+                ~table["$padj_column"].isna() &
+                (pd.to_numeric(table["$logFC_column"], errors='coerce').abs() >= float(logFC_threshold)) &
+                (pd.to_numeric(table["$padj_column"], errors='coerce') <= float("$padj_threshold"))]
+
+    table.to_csv(path.splitext(path.basename("$input_file"))[0]+"_filtered.tsv", sep="\t", index=False)
+
+    with open('versions.yml', 'a') as version_file:
+        version_file.write('"${task.process}":' + "\\n")
+        version_file.write("    python: " + (platform.python_version()) + "\\n")
```
pinin4fjords, line 50: Don't include python in the versions.yml unless you've pinned it in your environment.yml. Once we get to nf-test we'll snapshot the versions.yml, and the snapshot will break when the Python version drifts.

pinin4fjords (changes requested):

modules/local/filter_difftable.nf
WackerO: Ah okay, will remove that

.nf-core.yml
```diff
@@ -1 +1,8 @@
 repository_type: pipeline
+lint:
+  nextflow_config:
+    - config_defaults:
+        - params.logo_file
```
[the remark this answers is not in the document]
WackerO: That's for fixing the linting error; otherwise nf-core lint will complain about the default paths to the files differing in nextflow.config vs nextflow_schema.json

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Changed`
 
+- [[#264](https://github.com/nf-core/differentialabundance/pull/264)] - Change FILTER_DIFFTABLE to python because AWK does not correctly filter reliablys ([@WackerO](https://github.com/WackerO), review by [@pinin4fjords](https://github.com/pinin4fjords))
```
pinin4fjords, line 34: Typo

pinin4fjords (approved):
Just a typo (and maybe make the actual PR name match what's in the changelog), otherwise good.

PR #282, opened 2024-07-26
Dev

bjlang:
Copying the matrix as feature_matrix seems to have made the inputs appear to differ between identical runs, thus preventing caching.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [[#280](https://github.com/nf-core/differentialabundance/pull/280)] - Bump shinyngs, fix contrasts passed to app creation ([@pinin4fjords](https://github.com/pinin4fjords), review by [@WackerO](https://github.com/WackerO))
 - [[#274](https://github.com/nf-core/differentialabundance/pull/274)] - Fix pagination on samples table ([@pinin4fjords](https://github.com/pinin4fjords), review by [@WackerO](https://github.com/WackerO))
 - [[#272](https://github.com/nf-core/differentialabundance/pull/272)] - Show >10 contrasts in report ([@pinin4fjords](https://github.com/pinin4fjords), review by [@WackerO](https://github.com/WackerO))
+- In order to improve resumability, create a copy of the matrix as an annotation file only if necessary.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


WackerO:
Hello there, I think the caching fails not because of a different md5sum but because the copied matrix has another file name. Not sure exactly how the nextflow resume functionality works, but I assume it takes file names into consideration.
In any case, we cannot just remove this line from the code. The file copying is actually necessary for the shinyngs validator to run.

If we want the resume to work in these cases, we need to find another solution. Closing this for now; feel free to re-open if you have another idea :)

bjlang:
I believe the problem is rather the different timestamps of creation.
That's why I suggest to skip copying, iff this was copied by the previous run, i.e. the file exists AND the content of the file would be identical to the copied one. 
A caveat I se is that the suggestion actually would still only allow resuming in case someone runs the pipeline multiple times consecutively on the same input matrix, e.g. `nextflow run main.nf -resume -profile test_nogtf,docker --outdir result_test`, but the cache will still become invalidated everytime one uses a different input matrix. I guess this could be tackled by actually using different file names, because currently the filename is actually always the same (matrix_as_anno.tsv).

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [[#280](https://github.com/nf-core/differentialabundance/pull/280)] - Bump shinyngs, fix contrasts passed to app creation ([@pinin4fjords](https://github.com/pinin4fjords), review by [@WackerO](https://github.com/WackerO))
 - [[#274](https://github.com/nf-core/differentialabundance/pull/274)] - Fix pagination on samples table ([@pinin4fjords](https://github.com/pinin4fjords), review by [@WackerO](https://github.com/WackerO))
 - [[#272](https://github.com/nf-core/differentialabundance/pull/272)] - Show >10 contrasts in report ([@pinin4fjords](https://github.com/pinin4fjords), review by [@WackerO](https://github.com/WackerO))
+- In order to improve resumability, create a copy of the matrix as an annotation file only if necessary.
```
WackerO:
```suggestion
- [[#282](https://github.com/nf-core/differentialabundance/pull/282)] - In order to improve resumability, create a copy of the matrix as an annotation file only if necessary; add original matrix file name to copy name ([@bjlang](https://github.com/bjlang), review by [@WackerO](https://github.com/WackerO))
```

WackerO:
@bjlang As already discussed on slack, I didn't understand the code change correctly initially; thanks for the contribution! I'm approving, please first commit the suggestion that I made to the changelog before merging 😄

WackerO (approved):

WackerO:
Thanks for the work, @bjlang!

PR #288, opened 2024-09-13
Dev ratio

bjlang:
## PR checklist


suzannejin:
- [x] just finished adding the updated local modules `propr/propr` `propr/propd` and `propr/grea`

modules/local/propr/grea/main.nf
```diff
@@ -0,0 +1,24 @@
+process PROPR_GREA {
```
pinin4fjords, line 1: Are these modules local just for testing purposes? Assume they'll be in nf-core?

subworkflows/local/differential/main.nf
```diff
@@ -0,0 +1,86 @@
+//
+// Perform differential analysis
+//
+include { PROPR_PROPD as PROPD } from "../../../modules/local/propr/propd/main.nf"
+include { DESEQ2_DIFFERENTIAL  } from '../../../modules/nf-core/deseq2/differential/main'
+
+workflow DIFFERENTIAL {
+    take:
+    ch_tools
```
pinin4fjords: I don't think these 'thematic' subworkflows should take `ch_tools`. The content of that channel is only partially relevant to each subworkflow. 

Instead, they should take the actual things used by the workflow (e.g. 'diff_method'). We can use a `multiMap` outside the subworkflows to create the relevant channels.

pinin4fjords: In that way, we can develop a 'differential' subworkflow that calls DESeq2, Limma, whatever on command from transparent inputs, and it will be generally useful and nf-core material. 

Left as a channel input you have to manipulate in this way, it's far less useful.

conf/modules.config
```diff
@@ -481,4 +481,36 @@ process {
             enabled: false
         ]
     }
+
+    withName: "PROPR"{
+        ext.args = { "${meta.args_cor}" == "null" ? '' : "${meta.args_cor}" }
```
pinin4fjords: There should probably be provision for specifying args from global config, as well as from the tool sheet.

pinin4fjords (commented):
See where you're going- good work. I think we should integrate this philosophy, if not its detail, early on, since this will allow us to break assumptions, e.g. that RNA-seq = DESeq2.

Some points:

1. The thematic subworkflows for differential etc need to be better 'encapsulated'. They shouldn't need to know anything about the structure of complex input channels, and should receive everything they need as straightforward, separate inputs. maps, multiMaps etc to accomplish that can happen in the calling workflows. 

2. I don't actually like the separation between the 'experimental' arm and the rest of the workflow, it's just going make the logic difficult to maintain. So, I think we will need to harmonise the new and existing logic. 

Conceptually, each of the current technology-wise profiles ([rnaseq](https://github.com/nf-core/differentialabundance/blob/master/conf/rnaseq.config), [maxquant](https://github.com/nf-core/differentialabundance/blob/master/conf/maxquant.config), [affy](https://github.com/nf-core/differentialabundance/blob/master/conf/affy.config)), is something like one of the lines in your tools_samplesheet.csv- it's a set of parameters that dictates what path to use in the workflow. 

I think, probably, that before adding the new modules we should start by taking the existing logic, and reformulating along these lines. I think this means:

1. Push existing 'differential' logic into new subworkflow
2. Push existing gene set logic into new subworkflow
3. Convert existing technology profiles to pathways 

This should be backwards-compatible with current logic. So:

1. Parameters in the tools sheet should be valid workflow-level parameters. So I should be able to say `--diff_method deseq2`, as well as use `diff_method` in the tool sheet. 
2. NOT specifying a `pathway` should simply lead to ch_tools being populated with a single map containing the default values of relevant parameters, dependent on `-profile` . Then logic can proceed in the same way as if `ch_tools` had been populated using `tools` and `pathway`. 
3. Pathway specified parameters should take highest priority if the same parameters are specified on the command line. 
4. There will be a default tool sheet with all valid pathways, documentation to stress that users change this file at their peril.
6. Each pathway will be annotated with applicable study types to make it easy for people with e.g. rnaseq data to avoid inappropriate pathways.
7. When I specify `study_type` with an explicit pathway, it will check that that pathway is present in the tool sheet, has been marked as compatible with the `study type` of and then proceed. e.g. maybe that could be `rnaseq_voom` once we have Voom in place.
9. When I specify study_type `rnaseq` with `all` as the pathway, it will run all pathways marked as compatible. 
10. Not specifying `study_type` will enable any specified pathway(s) to run.

That should be easy enough to do, and that will be a basis on which

modules/local/propr/grea/main.nf
suzannejin: Exactly! For the moment, I want them there to try out things, before updating their final version for nf-core.
The reason for this, is that there are still some details that need to be decided (that will depend on the coordination with the rest of modules), and it would be an overkill to update to nf-core each time i change some minor stuff.

subworkflows/local/differential/main.nf
suzannejin: If I understand correctly your point, you want to parse the 'diff_method' and 'args_diff' meta values into one 'diff' channel?
That is actually what we aim to do with the next changes for 'ch_tools'. 
Currently, all the metas for different subworkflows are parsed inside 'ch_tools' and provided to each subworkflow.
But we want to modify it so that, 
   - differential subworkflow will only be provided with a ch_tools that see the metas 'diff_method' and 'args_diff'.
   - correlation subworkflow will only have a ch_tools that see 'cor_method' and 'args_cor'.
   - enrichment subworkflow will have a ch_tools that see everything ('diff_method', 'args_diff', 'cor_method', 'args_cor', 'enr_method', 'args_enr'). The reason is that the input coming from upstream will be different and they are characterized by their corresponding 'diff_method' 'args_diff' or 'cor_method' 'args_cor'.

The advantage to still have a 'ch_tools' inside the subworkflow are many:
  - we can easily use the information from 'ch_tools' to run one or more processes inside the same subworkflow.
  - We can parse the 'ch_tools' with its corresponding input data (ch_counts, ch_contrast, etc) in the way how each module requires, just before running the module. So it is visually clearer to understand what kind of parsing, etc are going on for a given module. I think in this way it is easier to understand what goes with which module, and hence also easier in the future to add new modules. For the new addition, one just have to consider to parse the inputs from the subworkflow with the channel just before calling this new module.

conf/modules.config
suzannejin: yes! I am currently working on this. It is allocated inside the other PR #302 , but still need to fix some stuff.

subworkflows/local/differential/main.nf
pinin4fjords: What I mean is, we **definitely** shouldn't have things like:

```
ch_tools
        .branch {
            propd:  it[0]["diff_method"] == "propd"
            deseq2: it[0]["diff_method"] == "deseq2"
        }
        .set { ch_tools_single }
```

That requires the subworkflow to know things about the calling suborkflow. The workflow should just `take:` the diff_method, and use that directly to condition what it does.

We can use `multiMap` outside of the subworkflow to generate the right channels:

```
ch_tools
    .multiMap { it ->
        ch_diff_method: it[0]["diff_method"]
        ch_other_option: it[0]["other_option"]
    }
    .set { result }
```

In the above example, result.ch_diff_method and result.ch_other option will maintain their relative order, so could be passed directly to DIFFERENTIAL. 

By doing things that way, the worklow has 'encapsulation'. You pass it the things it needs, and it doesn't need to know anything about what's done outside of the subworkflow. It's also portable, so it can become an nf-core subworkflow when it matures.

suzannejin:
- [x] added default parameters for propd/propr/grea modules. Still need to find a better way to handle them though, see discussion in #302

subworkflows/local/differential/main.nf
```diff
@@ -0,0 +1,173 @@
...
+include { DESEQ2_DIFFERENTIAL as DESEQ2_NORM } from "../../../modules/nf-core/deseq2/differential/main"
+include { LIMMA_DIFFERENTIAL } from '../../../modules/nf-core/limma/differential/main'
+include { FILTER_DIFFTABLE as FILTER_DIFFTABLE_LIMMA } from '../../../modules/local/filter_difftable'
+include { FILTER_DIFFTABLE as FILTER_DIFFTABLE_DESEQ2 } from '../../../modules/local/filter_difftable'
+
+workflow DIFFERENTIAL {
+    take:
+    ch_counts             // [ meta_exp, counts ] with meta keys: method, args_diff
+    ch_samplesheet        // [ meta_exp, samplesheet ]
+    ch_contrasts          // [ meta_contrast, contrast_variable, reference, target ]
```
pinin4fjords, line 15:
```suggestion
    ch_contrasts          // [ meta_contrast, contrast_variable, reference, target ]
    method                // limma, deseq2 ...
```

subworkflows/local/differential/main.nf
```diff
@@ -0,0 +1,173 @@
...
+    ch_adjacency                 = Channel.empty()
+
+    // branch tools to select the correct differential analysis method
+    ch_counts
+        .branch {
+            propd:  it[0]["method"] == "propd"
+            deseq2: it[0]["method"] == "deseq2"
+            limma:  it[0]["method"] == "limma"
+        }
+        .set { ch_counts }
```
pinin4fjords, line 36:
```suggestion
```

Sorry, I still don't like this. We should keep the interface simple rather than forcing people to stuff things into meta maps.

subworkflows/local/differential/main.nf
```diff
@@ -0,0 +1,173 @@
...
+include { PROPR_PROPD as PROPD } from "../../../modules/local/propr/propd/main.nf"
+include { DESEQ2_DIFFERENTIAL } from '../../../modules/nf-core/deseq2/differential/main'
+include { DESEQ2_DIFFERENTIAL as DESEQ2_NORM } from "../../../modules/nf-core/deseq2/differential/main"
+include { LIMMA_DIFFERENTIAL } from '../../../modules/nf-core/limma/differential/main'
+include { FILTER_DIFFTABLE as FILTER_DIFFTABLE_LIMMA } from '../../../modules/local/filter_difftable'
+include { FILTER_DIFFTABLE as FILTER_DIFFTABLE_DESEQ2 } from '../../../modules/local/filter_difftable'
+
+workflow DIFFERENTIAL {
+    take:
+    ch_counts             // [ meta_exp, counts ] with meta keys: method, args_diff
```
pinin4fjords, line 13:
```suggestion
    ch_abundance             // [ meta_exp, counts ] with meta keys: method, args_diff
```
This won't always be counts- e.g. for arrays it will be 'intensities', for PROTEUS it will be whatever that techology uses.

subworkflows/local/differential/main.nf
```diff
@@ -0,0 +1,173 @@
...
+            deseq2: it[0]["method"] == "deseq2"
+            limma:  it[0]["method"] == "limma"
+        }
+        .set { ch_counts }
+
+    // ----------------------------------------------------
+    // Perform differential analysis with propd
+    // ----------------------------------------------------
+
+    // TODO propd currently don't support blocking, so we should not run propd with same contrast_variable, reference and target,
```
pinin4fjords, line 42: You can assume that will be resolved at the workflow level (filtering the contrasts channel)

subworkflows/local/differential/main.nf
```diff
@@ -0,0 +1,173 @@
+//
+// Perform differential analysis
+//
+include { PROPR_PROPD as PROPD } from "../../../modules/local/propr/propd/main.nf"
+include { DESEQ2_DIFFERENTIAL } from '../../../modules/nf-core/deseq2/differential/main'
+include { DESEQ2_DIFFERENTIAL as DESEQ2_NORM } from "../../../modules/nf-core/deseq2/differential/main"
+include { LIMMA_DIFFERENTIAL } from '../../../modules/nf-core/limma/differential/main'
+include { FILTER_DIFFTABLE as FILTER_DIFFTABLE_LIMMA } from '../../../modules/local/filter_difftable'
```
pinin4fjords, line 8: This subworkflow will never run BOTH Limma AND DESeq2, so you don't need to import and call it twice. You should output DESeq2 + Limma to the same channels, which get filtered by the same process.

subworkflows/local/differential/main.nf
```diff
@@ -0,0 +1,173 @@
...
+    ch_adjacency                 = PROPD.out.adjacency.mix(ch_adjacency)
+
+    // ----------------------------------------------------
+    // Perform differential analysis with DESeq2
+    // ----------------------------------------------------
+
+    if (params.transcript_length_matrix) { ch_transcript_lengths = Channel.of([ exp_meta, file(params.transcript_length_matrix, checkIfExists: true)]).first() } else { ch_transcript_lengths = Channel.of([[],[]]) }
+    if (params.control_features) { ch_control_features = Channel.of([ exp_meta, file(params.control_features, checkIfExists: true)]).first() } else { ch_control_features = Channel.of([[],[]]) }
+
+    ch_counts
```
pinin4fjords: I don't think this logic should live in this subworkflow.

As far as the subworkflow is concerned, it should receive a single sample sheet, abundance matrix, length matrix etc, and set of contrasts. Complexity like this belongs at the workflow level.

pinin4fjords (changes requested):
I think we may have a difference of opinion here, but I **really** think components should be kept simple, and complexity around channel manipulations should be pushed up to workflows, it's just the nf-core way. 

I think I'm going to have a go at pushing the differential workflow up into nf-core/modules (in simplified form).

modules/local/propr/propd/main.nf
```diff
@@ -0,0 +1,33 @@
...
+    tag "$meta.id"
+    label 'process_medium'
+
+    // conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/bioconductor-limma_r-ggplot2_r-propr:209490acb0e524e3' :
+        'community.wave.seqera.io/library/bioconductor-limma_r-ggplot2_r-propr:17abd3f137436739' }"
+
+    input:
+    tuple val(meta), path(count), path(samplesheet), val(contrast_variable), val(reference), val(target)
```
pinin4fjords, line 11: This is very un-nf-core, and doesn't match the analagous differential modules. I think this will need to be split for nf-core

modules/local/propr/propd/main.nf
```diff
@@ -0,0 +1,33 @@
+process PROPR_PROPD {
+    tag "$meta.id"
+    label 'process_medium'
+
+    // conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/bioconductor-limma_r-ggplot2_r-propr:209490acb0e524e3' :
```
pinin4fjords, line 7: We need to use the https versions of the singularity URIs for Seqera Containers. It's a bit of a pain, but here's how: https://nfcore.slack.com/archives/CJRH30T6V/p1729023924708159?thread_ts=1729023915.530259&cid=CJRH30T6V

subworkflows/local/differential/main.nf
suzannejin: but that is the idea with the multiply running options no?
If the user wants to run --pathway deseq2_gsea,limma_gsea
Both deseq2 and limma would be run, but the output would be channeled into different places.
How would you deal with that instead?

modules/local/propr/propd/main.nf
suzannejin: okay that is not a problem

suzannejin: even though why this is not recommended by nf-core?

pinin4fjords: Ahh- has there been a change? Could you point at that guideline?

suzannejin: not sure if there a change/guideline for that...that is why i asked because i was not sure why you say this is un-nf-core.

subworkflows/local/differential/main.nf
pinin4fjords: Ahh yes, I think maybe this is down to our difference in thoughts on architecture.

In my view the subworkflow would run twice to make that happen.

modules/local/propr/propd/main.nf
pinin4fjords: I mean, the convention in modules is not to combine everything into a single input channel, but only to combine related things. You could copy the interface of the other differential modules, which would have better consistency and would prevent us having to do special operations to fit this module.

subworkflows/local/differential/main.nf
suzannejin: ahhhhh so you would have import various subworkflows (eg DIFFERENTIAL_DESEQ2, DIFFERENTIAL_LIMMA) inside the workflow and run them.... actually that is nice :)

pinin4fjords: Yep. We keep all components as simple as possible, and manage complexity via channel operations in the calling workflow.

subworkflows/local/differential/main.nf
bjlang: I don't see your problem. The meta is existing in every nf-core module and subworkflow and from what I see it is also not uncommon to use the meta information for downstream analysis decisions. Take for example the single_end flag in RNAseq, ChIPseq, ATACseq,... Both modules (e.g. [trimgalore](https://github.com/nf-core/chipseq/blob/2.1.0/modules/nf-core/trimgalore/main.nf) ) and [subworkflows](https://github.com/nf-core/chipseq/blob/76e2382b6d443db4dc2396e6831d1243256d80b0/subworkflows/local/bam_filter_bamtools.nf#L37](https://github.com/nf-core/chipseq/blob/76e2382b6d443db4dc2396e6831d1243256d80b0/subworkflows/nf-core/fastq_fastqc_umitools_trimgalore/main.nf#L60) ) are using it, and thus forcing the flag to exist. This is even also used for [branching](https://github.com/nf-core/chipseq/blob/76e2382b6d443db4dc2396e6831d1243256d80b0/subworkflows/local/bam_filter_bamtools.nf#L37).

pinin4fjords: The single-end thing is a special case, we try and discourage the propagation of assumptions about fields in the meta.

The branching you point at is necessitated by the need to direct outputs form the first processes to the right place. It's not engineered in to the inputs. If I were writing that subworkflow without that first process I'd be saying the same thing- the single end status would be a straighforward subworkflow input.

pinin4fjords: The point here is that the tool status has nothing to do with the count file. You're forcing people to build it into the meta to suit the use case, but it would just make life harder for other users of the subworkflow.

pinin4fjords: https://nf-co.re/docs/guidelines/components/modules#types-of-meta-fields

subworkflows/local/differential/main.nf
bjlang: I agree that access to params can be moved upstream in favour of the workflow providing these channels

suzannejin:
close pr as it is treated in #431 and #385

PR #302, opened 2024-10-11
[dev-ratio] add default parameters for propd/propr/grea

suzannejin:
- [x] added default parameters for propd/propr/grea into nextflow.config and schema
- [x] modified PROPD to consider the input from contrast file 

## PR checklist


conf/modules.config
```diff
@@ -482,35 +482,102 @@ process {
         ]
     }
 
+    // TODO
+    // for the moment, the parameters from toolsheet are parsed here.
+    // For the repeated arguments, since the R argument parser will only take the first one,
```
WackerO, line 487: What does 'repeated arguments' mean?

modules/local/propr/grea/main.nf
```diff
@@ -3,9 +3,9 @@ process PROPR_GREA {
     label 'process_high'
 
     // conda "${moduleDir}/environment.yml"
```
WackerO, line 5: Why is conda not an option for this module?

modules/local/propr/propd/templates/propd.R
```diff
@@ -323,8 +458,14 @@ if (opt\$permutation == 0) {
...
-        connectivity <- get_connectivity(adj)
-        hub_genes <- get_hub_genes(connectivity)
+        connectivity <- get_connectivity(
+            pd,
+            adj,
+            de,
+            theta_cutoff,
+            features_id_col=opt\$features_id_col
+        )
+        hub_genes <- get_hub_genes(connectivity, weigthted=opt\$weighted_degree)
```
WackerO:
```suggestion
        hub_genes <- get_hub_genes(connectivity, weighted=opt\$weighted_degree)
```

modules/local/propr/propr/main.nf
```diff
@@ -2,10 +2,10 @@ process PROPR_PROPR {
     tag "$meta.id"
     label 'process_medium'
 
-    conda "${moduleDir}/environment.yml"
+    // conda "${moduleDir}/environment.yml"
```
WackerO, line 5: Why no conda?

nextflow_schema.json
```diff
@@ -752,6 +750,97 @@
             },
             "fa_icon": "fas fa-border-all"
         },
+        "propd_specific_options": {
+            "title": "propd specific options",
+            "type": "object",
+            "description": "Parameters to run differential proportionality",
+            "default": "",
+            "properties": {
+                "propd_group_col": {
```
WackerO: Please add descriptions to this param and the following

modules/local/propr/grea/main.nf
suzannejin: so the module is temporal yet, 
I plan to finish the nf-corization of the module during the hackathon, and by then the conda environment and all the nf-tests should be ready

conf/modules.config
suzannejin: So for example, in the current samplesheet we have these pathways:
```
pathway_name,diff_method,args_diff,cor_method,args_cor,enr_method,args_enr
pcorbshrink,,,propr,--metric pcor.bshrink,,
propr,,,propr,--metric rho,,
cor,,,propr,--metric cor,,
```

The `metric` argument is changed depending on the pathway. Ideally, the default argument for metric declared in modules.config should be replaced. I imagine this can be done maybe by a bit of groovy scripting, but for the moment I don't know how to do it yet. So, my temporal workaround is to just append it to the ext.arg, this will create repeated arguments (eg. `--arg1 val1 --metric cor --metric rho`).  In R, this is parsed as a named list, and only the first one with name 'metric' will be considered, succesfully overwriting the behaviour.
This is not ideal, so I put the TODO, for later deal on it.

WackerO: Ah, I see. I'm wondering if for this it wouldn't be possible to make an exception to the nf-core rules and put these arguments into a channel. That way, the module would only receive these arguments once per module execution. Might need to discuss this with the core members.

WackerO (approved):

conf/modules.config
suzannejin: Yes we could further discuss this! I would like to hear in which way you would put them into a channel?
What I thought it could work is to run a groovy function above the ext args, that loads everything as a dict and replaces the repeated arguments, and return the parsed arguments as ext.arg for the execution of the module. 
But yes, it should have better alternative ways.

WackerO: Ah sorry, I thought I had replied here...
I'm not sure how exactly the code would have to be written, but basically I think it should be possible to create channels for each of the params (or maybe even just a single channel for all of them, similar to the meta map) that contain the settings of all params for a single pathway name, e.g.:
```
First iteration of the pipeline: [pathway_name: pcorbshrink, diff_method: ???, metric: pcor.bshrink, ... ]
Second iteration of the pipeline: [pathway_name: propr, diff_method: ???, metric: rho, ... ]
Third iteration of the pipeline: [pathway_name: cor, diff_method: ???, metric: cor, ... ]
```
The brackets all refer to the same channel, just different 'constellations' or whatever that is called in nextflow talk, i.e. when the contents of a channel change between different pathways of the same pipeline run.
Then we could just give this channel to all the modules that require this information, and in each pathway, the correct values are handed over to these modules :)

PR #304, opened 2024-10-21
Disallow TXT files

WackerO:
Currently, the pipeline does not properly handle contrast files with the TXT extension; they are treated as comma-separated which results in an error. This behaviour is now fixed.

## PR checklist


pinin4fjords (changes requested):
I don't agree with this, you've just flipped the problem and created another one.

txt doesn't imply any particular separator- commas or tabs. Therefore we should just restrict the input to csv|tsv and disallow txt - not assume that txt has tabs.

WackerO:
You're right, txt doesn't imply anything, but we have treated txt as a tab-separated file format for all the other files; your `read_delim_flexible` function makes the same assumption. I do think that we should treat this uniformly and either continue the same assumption here or disallow txt files in general.

pinin4fjords:
Yes, let's disallow!

pinin4fjords (approved):

PR #306, opened 2024-10-22
[dev-ratio] add deseq2 and gprofiler2 into the experimental branch

suzannejin:
## PR checklist


conf/modules.config
```diff
@@ -490,7 +490,7 @@ process {
     // However, later we need to better handle this, maybe by a bit of groovy scripting to
     // overwrite the repeated parameters (?)
 
-    withName: "PROPR"{
+    withName: PROPR {
```
bjlang, line 493:
```suggestion
    withName: 'PROPR' {
```

bjlang: For consistency I'd use ' throughout the config

conf/modules.config
```diff
@@ -580,4 +607,76 @@ process {
...
+        publishDir = [
+            path: { [
+                "${params.outdir}/",
+                "enrichment_analysis/",
+                "grea",
+                (meta.args_enr ? "-${meta.args_enr.replace('--','').replace(' ', '_')}" : ''),
+                (meta.diff_method ? "-from-${meta.diff_method}" : ''),
+                (meta.args_diff ? "-${meta.args_diff.replace('--','').replace(' ', '_')}" : ''),
+                (meta.cor_method ? "-from-${meta.cor_method}" : ''),
+                (meta.args_cor ? "-${meta.args_cor.replace('--','').replace(' ', '_')}" : ''),
```
bjlang, line 632: With splitting the toolsheet information into only the subworkflow relevant chunks, the data accessed here wouldn't be existing anymore. I'm not sure such an elaborate file naming is necessary in this case - could it not be reduced to the pathway name and args_enr? (I dont see a reason for 2 different pathways with identical args_diff, args_cor and args_enr to exist)

conf/test_experimental.config
```diff
@@ -29,10 +29,13 @@ params {
     input = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.samplesheet.csv'
     matrix =  'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.salmon.merged.gene_counts.top1000cov.tsv'
     contrasts =  'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.contrasts.csv'
-    pathway = "propd,propd_grea,propr,cor"
+
+    // analysis
+    pathway = "propd,propd_grea,propd_ora,propr,cor,deseq2_ora"
```
bjlang, line 34: You think propd_fdr and pcorbshrink can be safely omitted from the test?

subworkflows/local/differential/main.nf
```diff
@@ -2,14 +2,17 @@
 // Perform differential analysis
 //
 include { PROPR_PROPD as PROPD } from "../../../modules/local/propr/propd/main.nf"
-include { DESEQ2_DIFFERENTIAL  } from '../../../modules/nf-core/deseq2/differential/main'
+include { DESEQ2_DIFFERENTIAL as DESEQ2 } from '../../../modules/nf-core/deseq2/differential/main'
+include { FILTER_DIFFTABLE as FILTER_DESEQ2 } from '../../../modules/local/filter_difftable'
```
bjlang, line 6: I personally prefer to use renaming only when necessary, i.e. when a module is included multiple times. That way you always directly know for nf-core modules in which folder the module code is located, without needing to check the include statement.

subworkflows/local/enrichment/main.nf
```diff
@@ -34,20 +50,37 @@ workflow ENRICHMENT {
...
+            gmt : [gene_sets]
+            background : [counts]
+        }
+        .set{ ch_enrichment_gprofiler2 }
+
+    GPROFILER2_GOST(
+        ch_enrichment_gprofiler2.de,
+        ch_enrichment_gprofiler2.gmt,
+        ch_enrichment_gprofiler2.background
+    )
```
bjlang, line 88: I don't see a reason for the multiMap here.

Why not something like
```suggestion
    GPROFILER2_GOST(
        ch_results_genewise_filtered.filter { it[0]["enr_method"] == "gprofiler2" },
        ch_gene_sets.first()[1],
        ch_counts.first()[1]
    )
```

subworkflows/local/experimental/main.nf
```diff
@@ -30,11 +30,26 @@ workflow EXPERIMENTAL {
...
+        ch_control_features = Channel.of([ [ "id": params.study_name  ], file(params.control_features, checkIfExists: true)]).first()
+    } else {
+        ch_control_features = [[],[]]
+    }
+    if (params.transcript_length_matrix) {
+        ch_transcript_lengths = Channel.of([ [ "id": params.study_name  ], file(params.transcript_length_matrix, checkIfExists: true)]).first()
+    } else {
+        ch_transcript_lengths = [[],[]]
+    }
+
```
bjlang, line 53: As this is Deseq2 specific, wouldn't it be better to move it into the differential_analysis subworkflow?

conf/modules.config
suzannejin: but if you have two pathways:
1. deseq2 + gprofiler2
2. deseq2 + gsea

Would you save same deseq2 output twice, but into two different pathway directories?

conf/test_experimental.config
suzannejin: propd_fdr is currently a bit useless for this test, because it is not giving any significant results for the test dataset, so it can be ignored.
But I would have to check other datasets for it to run and test

Then instead of pcorbshrink maybe I would ignore cor

subworkflows/local/experimental/main.nf
suzannejin: it is good to call params within the subworkflows? I thought that should be avoided for a better encapsulation of the subworflow?
Also, not now, but in the future, I think it make sense to have two separated blocks with DESEQ2_NORM dealing the normalization of the data (so in the data processing block) that will use these control/transcript length features, and DESEQ2_DIFFERENTIAL running the differential analysis on the already normalized data

conf/modules.config
bjlang: I'm talking only about the enrichment methods like the GREA process here, where you would access not only enr arguments but also need to pass over all other args from previous steps to build the filename in the way you suggest. So, what I mean is to use only args_enr and if this is not specific enough maybe additionally the pathway_name.
For deseq2 in the differential analysis step using args_diff to build the name is fine.

subworkflows/local/experimental/main.nf
bjlang: Yes I see your point, but I would argue that we have a better encapsulation, when a subworkflow can be so transparently changed (i.e. by adding a new module) that there is no need for making any changes to the calling workflow.

conf/modules.config
suzannejin: but now that we don't have the pathway name info inside the meta anymore, we don't have any way to distinguish between them and store them into different places (?)

Issue #340, opened 2024-11-07
Volcano plot in report shows less genes than in 'plots/differential/*/png/volcano.png'


d4straub:
### Description of the bug

The volcano plots in `results/report/study.html#Differential_analysis` differ to the ones in `results/plots/differential/*/png/volcano.png`, screenshots below.

I strongly assume its caused by the gtf I used. I downloaded gtfs from NCBI for non-standard organisms (bacteria). While I acknowledge that the gtfs might not follow standards (even if its coming from a popular resource) it is worrisome that the figure differ without any warning or error.

I havent investigated the issue, but I assume its related to https://github.com/qbic-pipelines/rnadeseq/issues/216. Most likely I will tackle that problem in the next days and might be able to point to the reason.

### Relevant files

`results/report/study.html#Differential_analysis`

![Image](https://github.com/user-attachments/assets/b715bcad-6e5d-4a55-86a4-e0cb2fd78ba3)

`results/plots/differential/*/png/volcano.png`

![Image](https://github.com/user-attachments/assets/dc5bee73-aa47-4d82-8c71-9324161f01b9)

### System information

Version of nf-core/differentialabundance 1.5.0

grst:
Seems like a duplicate of https://github.com/nf-core/differentialabundance/issues/270

d4straub:
I respectfully disagree. 
- I do have an overlapping problem: colors are missing and no labels -> amending the gtf with `gene_name` alleviates that
- New problem: most genes are even not listed in the report.

So about the latter:
head of `tables/differential/non-targeting_vs_control.deseq2.results_filtered.tsv` -> 3 upregulated genes (highlighted)
![Image](https://github.com/user-attachments/assets/229f9430-806f-48dc-8ce1-71e999b2e8a6)
the plot in `plots/differential/non-targeting_vs_control/png/volcano.png` -> 3 upregulated genes (all good!)
![Image](https://github.com/user-attachments/assets/6fd80139-b364-4534-9fd8-4d6bc24cf2e2)
there is no table for `Non-Targeting versus Control in Treatment` in report/study.html, instead:
```
No significantly differential ‘up’ genes.
No significantly differential ‘down’ genes.
```
and additionally the volcono plot shows no significant genes:
![Image](https://github.com/user-attachments/assets/aeb8143d-223e-4a38-b2cc-a7af2a5d3836)

Any idea? Did I miss that problem in the other issue? If so, sorry for opening it again.

Edit - my struggle:
[...]

PR #354, opened 2024-11-14
Add a warning in the RMD report when some rows are lost after merging Diff and Fatures tables

alanmmobbs93:
# PR checklist

- [X] This comment contains a description of changes (with reason).
- [X] `CHANGELOG.md` is updated.

# Issue reported:

On [#340](https://github.com/nf-core/differentialabundance/issues/340) it was reported that many genes were lost in the volcano plot generated by the R Markdown report, but that the PNG file was correct. These genes get lost if the annotation table doesn't contain them, resulting in reduced plots and results in this last part. As mentioned by the user, big discrepancies can be easily catched, but the small ones can go unnoticed, ending up in undesired results.

So, a warning message was added if the annotated differential expressed table contains fewer rows. Other solutions can be:
- Retain all rows even if there's no data about these genes.
- Report the error and stop the pipeline (I believe this should be done as an early validation when starting the pipeline).

### DIFF
--- a/assets/differentialabundance_report.Rmd
+++ b/assets/differentialabundance_report.Rmd
@@ -399,8 +399,6 @@ results <- lapply(differential_files, function(diff_file) {
 
         # Merge tables
         diff_features <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column)
-        # test
-        diff_features <- diff_features[1:50,]
 
         # Get number of rows before and after merging
         length_diff <- as.numeric(nrow(diff))
--- a/assets/differentialabundance_report.Rmd
+++ b/assets/differentialabundance_report.Rmd
@@ -370,10 +370,17 @@ differential_files <- lapply(contrasts$id, function(d){
     file.path(params$input_dir, paste0(gsub(' |;', '_', d), differential_file_suffix))
 })
 
-differential_results <- lapply(differential_files, function(diff_file){
-    if (! file.exists(diff_file)){
+# Initialize vector to store warning messages before merging tables
+warnings_list <- c()
+
+# Read differential results and merge with features table
+results <- lapply(differential_files, function(diff_file) {
+    warnings <- c()  # Initialize local warning vector
+
+    if (!file.exists(diff_file)) {
         stop(paste("Differential file", diff_file, "does not exist"))
     }
+
     diff <- read_differential(
         diff_file,
         feature_id_column = params$differential_feature_id_column,
@@ -382,19 +389,43 @@ differential_results <- lapply(differential_files, function(diff_file){
         qval_column = params$differential_qval_column
     )
 
-    # If fold changes are not logged already, log them (we assume they're logged
-    # later on)
-
-    if (! params$differential_foldchanges_logged){
+    # If fold changes are not logged already, log them
+    if (!params$differential_foldchanges_logged) {
         diff[[params$differential_fc_column]] <- log2(diff[[params$differential_fc_column]])
     }
 
     # Annotate differential tables if possible
-    if (! is.null(params$features)){
-        diff <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column)
+    if (!is.null(params$features)) {
+
+        # Merge tables
+        diff_features <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column)
+
+        # Get number of rows before and after merging
+        length_diff <- as.numeric(nrow(diff))
+        length_features <- as.numeric(nrow(diff_features))
+
+        # Compare numbers and report
+        if (length_diff != length_features) {
+            warnings <- c(warnings,
+                paste0(
+                    '<p style="color:red;"><strong>WARNING:</strong> Some rows from the differential expressed table (', basename(diff_file), ') were lost after merging with features table (', basename(params$features), ').\n',
+                    'Rows in diff table: ', length_diff, '.\n',
+                    'Rows in merged table: ', length_features, '.</p>\n'
+                )
+            )
+        }
+    } else {
+        diff_features <- diff
     }
-    diff
+
+    # Return both the results and the local warnings
+    list(diff_features = diff_features, warnings = warnings)
 })
+
+# Separate differential_results and warnings_list from results
+differential_results <- lapply(results, `[[`, "diff_features")
+warnings_list <- unlist(lapply(results, `[[`, "warnings"))
+
 names(differential_results) <- contrasts$id
 ```
 
@@ -787,7 +818,6 @@ foo <- lapply(names(p_value_types), function(pvt){
 ```
 
 ```{r, echo=FALSE, results='asis', eval = FALSE}
-
 differential_summary_string <- paste(
     paste(
     lapply(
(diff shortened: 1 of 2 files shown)


pinin4fjords:
A lot of other changes here, do you need to sort out via Git?

alanmmobbs93:
@pinin4fjords Sorry about that, it's corrected now

assets/differentialabundance_report.Rmd
```diff
@@ -382,19 +389,43 @@ differential_results <- lapply(differential_files, function(diff_file){
...
         diff[[params$differential_fc_column]] <- log2(diff[[params$differential_fc_column]])
     }
 
     # Annotate differential tables if possible
-    if (! is.null(params$features)){
-        diff <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column)
+    if (!is.null(params$features)) {
+
+        # Merge tables
+        diff_features <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column)
```
d4straub: The warnings are great. But I am wondering what would be the consequence of
```suggestion
        diff_features <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column, all.x = FALSE, all.y = TRUE)
```
would that lead to a complete plot (all features present) but with NA for (some) annotations? If so, that might be a nice solution as well.

alanmmobbs93: @d4straub As I can see it, the volcano plots won''t be affected, but I think this should be reported anyway (number of missing rows or number of features with NAs). We might take care of the new plots recently added below too (still in dev). I'm not sure if they can handle tables with NAs or without these columns at all @atrigila

assets/differentialabundance_report.Rmd
```diff
@@ -382,19 +389,43 @@ differential_results <- lapply(differential_files, function(diff_file){
...
+
+        # Get number of rows before and after merging
+        length_diff <- as.numeric(nrow(diff))
+        length_features <- as.numeric(nrow(diff_features))
+
+        # Compare numbers and report
+        if (length_diff != length_features) {
+            warnings <- c(warnings,
+                paste0(
+                    '<p style="color:#DAA520;"><strong>WARNING:</strong> Some rows from the differential expressed table (', basename(diff_file), ') were lost after merging with features table (', basename(params$features), ').\n',
```
pinin4fjords:
```suggestion
                    '<p style="color:#DAA520;"><strong>WARNING:</strong> Some rows from the differential expressed table (', basename(diff_file), ') were absent from the features table (', basename(params$features), ') and lost on merge.\n',
```

assets/differentialabundance_report.Rmd
```diff
@@ -382,19 +389,43 @@ differential_results <- lapply(differential_files, function(diff_file){
...
+
+        # Merge tables
+        diff_features <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column)
+
+        # Get number of rows before and after merging
+        length_diff <- as.numeric(nrow(diff))
+        length_features <- as.numeric(nrow(diff_features))
+
+        # Compare numbers and report
+        if (length_diff != length_features) {
```
pinin4fjords: Not actually sure this is the right comparison. Maybe it would be better to check the dimensions of diff before and after the merge? Maybe also distinguish the possibility of there being MORE rows after the merge.

pinin4fjords (changes requested):
Some minor comments, but we should probably handle this upstream actually. We should be able to guarantee that this step receives a feature table with all the features. I feel like if that's not happening there's a bigger problem elsewhere.

assets/differentialabundance_report.Rmd
alanmmobbs93: @pinin4fjords I like the idea of adding the possibility of more rows, it would mean duplications, right? Working on that! But I would still compare only the number of rows, the dimensions (specifically the number of columns) will change based on the input parameters

alanmmobbs93: @pinin4fjords I agree that this should be addressed really upstream (as mentioned on the PR description), but the rmd code should contain the error handling as well, in case anyone runs it alone. What do you think?

pinin4fjords: Yeah, fair enough.

alanmmobbs93: @pinin4fjords Code updated, I changed a little bit the conditions to be more secure. It was true dimensions alone were not sufficient

atrigila:
Could you merge the latest changes from dev? I've incorporated some modifications to the `ci.yml`, that should make your tests run :).

pinin4fjords (commented):
The new logic is quite verbose, I asked ChatGPT to suggest a more concise version:

```
# Read differential results and merge with features table
results <- lapply(differential_files, function(diff_file) {
    if (!file.exists(diff_file)) stop(paste("Differential file", diff_file, "does not exist"))

    diff <- read_differential(
        diff_file,
        feature_id_column = params$differential_feature_id_column,
        fc_column = params$differential_fc_column,
        pval_column = params$differential_pval_column,
        qval_column = params$differential_qval_column
    )

    # Log transform fold changes if not already logged
    if (!params$differential_foldchanges_logged) {
        diff[[params$differential_fc_column]] <- log2(diff[[params$differential_fc_column]])
    }

    # Annotate differential table if features table is provided
    if (!is.null(params$features)) {
        merged <- merge(features, diff, by.x = params$features_id_col, by.y = params$differential_feature_id_column)
        n_missing <- length(setdiff(diff[[params$differential_feature_id_column]], merged[[params$features_id_col]]))
        warnings <- c(
            if (n_missing > 0) sprintf(
                '<p style="color:#DAA520;"><strong>WARNING:</strong> %d IDs from the differential table (%s) were lost on merge with features table (%s).</p>',
                n_missing, basename(diff_file), basename(params$features)
            ),
            if (nrow(merged) < nrow(diff)) sprintf(
                '<p style="color:#DAA520;"><strong>WARNING:</strong> Rows were lost on merge (%s -> %s). Original: %d, Merged: %d.</p>',
                basename(diff_file), basename(params$features), nrow(diff), nrow(merged)
            ),
            if (nrow(merged) > nrow(diff)) sprintf(
                '<p style="color:#DAA520;"><strong>WARNING:</strong> Rows were duplicated on merge (%s -> %s). Original: %d, Merged: %d.</p>',
                basename(diff_file), basename(params$features), nrow(diff), nrow(merged)
            )
        )
    } else {
        merged <- diff
        warnings <- character(0)
    }

    list(diff_features = merged, warnings = warnings)
})

# Separate differential_results and warnings_list from results
differential_results <- lapply(results, `[[`, "diff_features")
warnings_list <- unlist(lapply(results, `[[`, "warnings"))
```

What do you think?

It would also be good to have a 'before and after' of the report from test_full, to make sure nothing is changed unexpectedly. I can do it, but not quickly, so if you have time to do that and post them here that would be awesome.

alanmmobbs93:
@pinin4fjords Done! I just added some comments to your solution to make it not only shorter but also readable. I cannot attach the output HTML files here, If you know how to do it just let me know. To my surprise, some genes were lost in the process when running the test_full profile

pinin4fjords:
> To my surprise, some genes were lost in the process when running the test_full profile

OK, if this is having an impact on results we'll need to spend some time figuring that out before any merge

alanmmobbs93:
Hello @pinin4fjords! I'd kindly suggest going on with this PR. This is a preexisting problem that got visualized now with these warnings (reinforcing the concept of always checking these steps). It's not about the new code as you can see with the following outputs:

Printing `dim(diff)` on dev branch 
```r
## Dimensions of DIFF before merging
## 31317 4 
## Dimensions of DIFF after merging
## 27743 6
```

Printing `dim(diff)` on feature branch:
```r
## Dimensions of DIFF before merging
## 31317 4 
## Dimensions of DIFF after merging
## 27743 6 
```

We can work on the issue after this

pinin4fjords (approved):
OK, sorry, misunderstood your comment, go ahead :-)

Issue #367, opened 2024-12-03
Meta issue functional analyses


grst:
### Description of feature

Functional analyses are often more easier to interpret than the raw list of differential gene expression analyses. 
Let's use this issue to get an overview of use-cases and methods we want to implement. 

It's a non-goal to add all possible methods for Gene Set Enrichment Analysis. Instead let's focus on 1-2 methods per use case that implement the current best practice and are computationally efficient. 

A general consideration here is if the methods should be executed on the DE results (e.g. GSEA, decoupleR on statistics, ORA ...) or 
if they should be executed on the gene expression (e.g. ssGSEA, decoupleR on TPM, ...). The former means the statistics are computed *on genes* while the latter means statistics are computed *on samples*. 

## Gene set enrichment analysis of custom gene sets and or predefined gene sets (e.g. GO, HALLMARK, ...)

- GSVA #24
- GSEA #311 + already implemented
- PLAGE #25
- SingScore #26
- GSEA preranked #36
- Over representation analysis (ORA) #83
- decoupleR #350 
- gProfiler (already implemented)

## Transcription factors

- collecTRI + decoupleR #99 #349 #352

## Cancer pathways

- PROGENy + decoupleR #99 #351

## cell type deconvolution

- immundeconv for immune cell signatures https://github.com/nf-core/differentialabundance/issues/361 #368 
- deconvolution with custom (often single-cell based) signatures: See https://www.biorxiv.org/content/10.1101/2024.06.10.598226v1 for benchmark of methods

## Reporting

- All results should come with an appropriate visualization and HTML report. 

Anyone feel free to suggest other databases and/or methods. 

CC @tschwarzl @apeltzer @atrigila @nschcolnicov @alanmmobbs93

grst:
Notably, decoupleR implements the following methods from the list: 
 * GSVA
 * ORA
 * GSEA

grst:
> A general consideration here is if the methods should be executed on the DE results (e.g. GSEA, decoupleR on statistics, ORA ...) or 
if they should be executed on the gene expression (e.g. ssGSEA, decoupleR on TPM, ...). The former means the statistics are computed *on genes* while the latter means statistics are computed *on samples*. 

For us, it would certainly be useful to compute signature scores per sample as we may need to report them to a clinical database. The per-sample scores should be independnet of other samples as we have no control over what subsets of the data may be retrieved by others. SingScore/decoupleR on TPM certainly fulfil these criteria. 

[...]

Issue #361, opened 2024-11-19
Create subworkflow for deconvolution analyses in bulk data


apeltzer:
### Description of feature

Celltype deconvolution could be a subworkflow in nf-core and then be imported (with some adjustments) into multiple pipelines, one of them being differentialabundance as something users could turn on / off on demand. 

This could encompass: 

- [x] #368
* omnideconv (?)
* gene set analysis via GSVA (not sure about overlaps with other functionality here already - but can be discussed with community)

x-ref @christopher-mohr same subworkflow would be interesting for e.g. nanostring.

PR #421, opened 2025-01-15
Add immunedeconv module

nschcolnicov:
## PR checklist


pinin4fjords:
sorry, I'm not familiar with immunedeconv. Could you please update the description to describe the module and how it fits into the overall themes of the workflow? The name makes it sound like something quite specialist.

grst:
Hi @pinin4fjords, see also the corresponding issue here: 
https://github.com/nf-core/differentialabundance/issues/361

Cell-type deconvolution is common downstream analysis step for RNA-seq data to estimate (immune) cell-type fractions and compare them between conditions. It would be one of several methods for "functional analyses" that we think this workflow could benefit from: https://github.com/nf-core/differentialabundance/issues/367

The one thing I'd like to see in addition to what's done in this PR is to include the results in some sort of report. But this can (and probably should) be addressed separately.

pinin4fjords:
OK, looking pretty good. I assume all the signatures are 'baked in' to this method- does there need to be any provision for a custom matrix to supply those?

nschcolnicov:
> OK, looking pretty good. I assume all the signatures are 'baked in' to this method- does there need to be any provision for a custom matrix to supply those?

Hi @pinin4fjords as far as I understand the tool, this is currently not possible: https://omnideconv.org/immunedeconv/articles/immunedeconv.html#can-i-specify-a-custom-signature-matrix-through-immunedeconv, maybe @grst can confirm.

grst:
`immundeconv` supports custom matrices for some tools, but that's more of a niche use-case. It's really intended to be ran with the predefined signature that are shipped with the methods. 

Having predefined signatures makes it really convenient to use. There's a lot of other methods that support custom signature matrices (typically derived from scRNA-seq data) that we collected in [omnideconv](https://github.com/omnideconv/omnideconv). 
Supporting such tools could be interesting, too, but that would be a separate module.

pinin4fjords (changes requested):
OK, gotcha.

I'm sorry, I hate to make you do it (mostly because I hate doing it), but I think you're going to have to update the metro map before this gets merged.

pinin4fjords (approved):

suzannejin:
Hey! a quick question, if I understood well, the test profile that runs deseq2+gsea would not interact at all with immunodev?
If so, why not create a separate test profile?

grst:
Correct, at least for now this is completely independent.

Issue #411, opened 2024-12-20
Merge dev_tmp branch and revert CI changes


nschcolnicov:
We created a temporary dev branch to work as a dev branch during the holidays and push all of the changes there. 
This branch will be reviewed and merged again in january, once everyone is back.
The ci.yml file needs to be updated so that tests now also run for the dev_tmp branch, I added a TODO to remember to revert this change once the branch is merged.
This ticket is created as a reminder of this.
FYI @grst @atrigila @alanmmobbs93 @apeltzer

nschcolnicov:
Created PR for this: https://github.com/nf-core/differentialabundance/pull/419 so we don't forget.

Next steps are:
- Reverting changes in ci.yml
- Reviewing changelog.md file

nschcolnicov:
Dev_tmp branch won't be merged, instead we will create separate PRs for each change. This is the summary of all the changes included in that branch:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/ns39399/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/ns39399/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">


<body link="#467886" vlink="#96607D">


PR | Description | Ticket | Status | Author | PR to dev | To be done by
-- | -- | -- | -- | -- | -- | --
https://github.com/nf-core/differentialabundance/pull/382 | Adding YAML contrasts file handling + schema validation | https://github.com/nf-core/differentialabundance/issues/405 | DONE | nschcolnicov | https://github.com/nf-core/differentialabundance/pull/410
  |   | https://github.com/nf-core/differentialabundance/issues/370 | DONE | nschcolnicov | https://github.com/nf-core/differentialabundance/pull/410
https://github.com/nf-core/differentialabundance/pull/404 | Implement model validation | https://github.com/nf-core/differentialabundance/issues/371 | To be added into dev | alanmmobbs93 | Not done yet | @atrigila  
[...]

PR #404, opened 2024-12-18
New Feature POC: VALIDATE_MODEL

alanmmobbs93:
## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core lint`).
- [x] Ensure the test suite passes (`nf-test test main.nf.test -profile test,docker`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

Related to: [#371](https://github.com/nf-core/differentialabundance/issues/371)

## NEW FEATURE: `VALIDATE_MODEL`
**GOAL** 
Create a validation step to cross information between the YML (models and contrast information) and the sample sheet.

#### Inputs:
 * YML
 * Samplesheet
 * sample_id_col
 
#### Outputs: 
 * Validated phenotypic table that contains only the columns that were required from the YML file
 * models.json that contains info about full rank models
 * Warnings JSON file, in case we'd like to report it in the workflow.

#### Functionalities:
 * All variables included in the YML file must be present in the sample sheet as column (first component of the contrast definitions, a better solution would be to detect it from the formula if we decide to keep it in the yml).
 * Blocking factors are also checked for existence.
 * All levels declared for the variables (extracted from `contrast` field) must be present in the column. If there are more levels in the samplesheet, they are reported with a warning.
 * Control special characters.
 * Control missing values.
 * Models are constructed with base R functions, and contrast are generated in order to check if the models will be full ranked or not. In case they are not, warnings are generated. We can decide whether to report this at nextflow level or not.

#### Perspectives:
 * Easy to include more fields if they are added to the yml.
 
### Testing
* I'd like to, and invite everyone, to test it with real cases to check the flexibility in reading variables from the YML file and finding real errors.
[...]

Issue #371, opened 2024-12-03
Model validation


grst:
### Description of feature

This issue is to follow up on the request in #362 to implement model validation. The goal is to catch any errors in the model definition/contrast specification as early as possible for a fast feedback loop. 

List of things to check (feel free to add items):

- [x] columns from design matrix are present in sample metadata
- [x] design matrix is full rank
- [x] contrasts are valid (i.e. specified coefficients occur as columns in the design matrix)
- [x] no NAs in any of the columns used in the design
- [x] warn about continuous covariates (I remember a case where numeric patient IDs were considered as a continuous covariate)
- [x] values of specified columns don't contain any special characters that the pipeline can't handle (e.g. `PR/CR` as a factor level fails downstream in clusterProfiler because of the `/`). 

## Implementation

Probably convenient to do it in R... but could we even do this in groovy directly? Then errors would be instant and we woudn't need to wait for a process to be fired up. 

CC @apeltzer @tschwarzl @atrigila @alanmmobbs93 @nschcolnicov

alanmmobbs93:
**POC**
Create a validation step to cross information between the YML (models and contrast information) and the sample sheet.

### Features:
#### Inputs:
 * YML
 * Samplesheet
 * sample_id_col
 
#### Outputs: 
 * Validated phenotypic table that contains only the columns that were required from the YML file
 * models.json that contains info about full rank models
 * Warnings JSON file, in case we'd like to report it in the workflow.

#### Functionalities:
 * All variables included in the YML file must be present in the sample sheet as column (first component of the contrast definitions, a better solution would be to detect it from the formula if we decide to keep it in the yml).
 * Blocking factors are also checked for existence.
 * All levels declared for the variables (extracted from `contrast` field) must be present in the column. If there are more levels in the samplesheet, they are reported with a warning.
 * Control special characters.
 * Control missing values.
 * Models are constructed with base R functions, and contrast are generated in order to check if the models will be full ranked or not. In case they are not, warnings are generated. We can decide whether to report this at nextflow level or not.

#### Perspectives:
 * Easy to include more fields if they are added to the yml.
 
### Testing
* I'd like to, and invite everyone, to test it with real cases to check the flexibility in reading variables from the YML file and finding real errors.
* Some errors should be found in the YML validation previous to this one.
* The local module and basic `nf-test` was also added for future changes and easy comparison during development.

### Test

[...]

PR #431, opened 2025-01-30
Integrate subworkflows into pipeline

suzannejin:
## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core lint`).
- [ ] Ensure the test suite passes (`nf-test test main.nf.test -profile test,docker`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

PR #436, opened 2025-02-14
Validate YML contrasts

atrigila:
The description of this new feature is here: https://github.com/nf-core/differentialabundance/issues/371
I am only merging it to dev as stated here: https://github.com/nf-core/differentialabundance/issues/411#issuecomment-2643221489.
I had to make a few adjustments of which parameters takes as input (`--contrasts_yml` instead of `--contrasts`), but other than those minor changes, this whole new feature was developed by @alanmmobbs93 here: https://github.com/nf-core/differentialabundance/pull/404.

## PR checklist


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [[#286](https://github.com/nf-core/differentialabundance/pull/286)] - Integration of limma voom for rnaseq data ([@KamilMaliszArdigen](https://github.com/KamilMaliszArdigen), review by [@pinin4fjords](https://github.com/pinin4fjords))
+- [[#380](https://github.com/nf-core/differentialabundance/pull/380)] - Replace local filter_diff_table module with nf-core one, and create nf-tests for tabular_to_gsea_chip. ([@nschcolnicov](https://github.com/nschcolnicov), review by [@pinin4fjords](https://github.com/pinin4fjords))
+- [[#382](https://github.com/nf-core/differentialabundance/pull/382)] - Add YAML formatted contrasts file handling. ([@nschcolnicov](https://github.com/nschcolnicov), review by [@TODO](https://github.com/TODO))
+- [[#411](https://github.com/nf-core/differentialabundance/pull/411)] - Replace local tabulartogseachip with nf-core version. ([@nschcolnicov](https://github.com/nschcolnicov), review by [@TODO](https://github.com/TODO))
```
grst: Is there any reason why this needs to be in one PR? If not I'd suggest to make a separate one with the module updates which should be rather non-controversal that we can merge first and then a second one with only the changes related to YAML validation.

suzannejin: the update of modules and subworkflows are also in #427
I would imagine it being merged in these couple of days, and then you won't need an separate PR

grst: https://github.com/nf-core/differentialabundance/pull/431 you mean? But fair point

modules/local/shinyngs/validatefomcomponents/main.nf
nschcolnicov, line 1: This module is not in use anymore, was only ever implemented in tmp_dev but the change was then reverted, we are using the nf-core one in the main dev branch. Please delete the local version of validatefomcomponents.

modules/local/shinyngs/validatefomcomponents/environment.yml
nschcolnicov, line 1: Please delete

modules/local/shinyngs/validatefomcomponents/meta.yml
nschcolnicov, line 1: Please delete

CHANGELOG.md
nschcolnicov: @grst @suzannejin The changelog is currently inaccurate as it contains the changes that were previously included in dev_tmp branch. 
The actual changes that are included in this PR are:
- Replace local tabulartogseachip module with nf-core one.
- Add model validation module.

@atrigila Please replace these three lines with a new one describing only these two changes and also update the PR number to #436

The only change that we could revert, since it was already going to be addressed in your PR @suzannejin is the replacement of tabulartogseachip module with the nf-core one, but it also doesn't hurt to have it here, I would recommend leaving it in this PR to avoid additional steps.

nschcolnicov (changes requested):
Looks good overall and tests are passing, some minor changes needed

CHANGELOG.md
atrigila: Thank you for your feedback, these suggestions have been addressed now.

modules/local/validatemodel/resources/usr/bin/validate_model.R.bkp
grst, line 1: can also be deleted I guess?

nextflow.config
```diff
@@ -389,6 +389,11 @@ process.shell = [
 // Disable process selector warnings by default. Use debug profile to enable warnings.
 nextflow.enable.configProcessNamesValidation = false
 
+// Enable loading modules bin files
+nextflow.enable.moduleBinaries = true
+
+def trace_timestamp = new java.util.Date().format( 'yyyy-MM-dd_HH-mm-ss')
```
grst: This change doesn't seem related? Is it here on purpose?

atrigila: Let me ask @alanmmobbs93 if it has some functionality, but I also think this should be deleted.

modules/local/validatemodel/resources/usr/bin/validate_model.R.bkp
atrigila: Let me ask @alanmmobbs93 if it has some functionality, but I also think this should be deleted.

atrigila: This can be safely deleted.

nextflow.config
atrigila: `trace_timestamp` was a line from the old template. I've removed it now.

nschcolnicov (approved):
Thanks for addressing the comments, LGTM

modules/local/validatemodel/resources/usr/bin/validate_model.R
```diff
@@ -0,0 +1,431 @@
...
+
+        # Populate contrasts_list for later model validation
+        contrasts_list[[CONTRAST$id]] <- list(
+            "variable" = variable,
+            "contrast" = levels,
+            "blocking_factors" = CONTRAST$blocking_factors
+        )
+
+        # Gather contrasts per variable into temporary storage
+        if (variable %in% names(temp_var)) {
```
suzannejin: i don't see where names(temp_var) is defined...?

modules/local/validatemodel/resources/usr/bin/validate_model.R
```diff
@@ -0,0 +1,431 @@
...
+## Print explicit messages
+for (INDEX in 1:length(var)) {
+    cat("\033[32mDetected '", names(var)[INDEX], "' variable with ", paste(var[[INDEX]], collapse = " "), " levels.\033[0m \n", sep = "")
+}
+
+# VALIDATE SAMPLESHEET BASED ON YML FILE ----------------------------
+## Define function to compare YML and samplesheet
+validate_model <- function(sample_column, variables, samplesheet) { # sample_column: string; variables: list; samplesheet: df
+
+    #######################################################
```
suzannejin: nice comments, they are super clear :)

modules/local/validatemodel/resources/usr/bin/validate_model.R
```diff
@@ -0,0 +1,431 @@
...
+            if ( is.numeric(samplesheet[[ VARIABLE ]]) ) {
+                continuous <- c(continuous, VARIABLE)
+            } else {
+                paste(VARIABLE, "not continuous")
+            }
+        }
+    }
+
+    ## Check variables and levels specified
+    for (VARIABLE in names(variables)[ names(variables) != "blocking_factors" ] ) { ## Exclude blocking factors, they cannot be treated equally here
```
suzannejin: maybe you could specify why they cannot?

modules/local/validatemodel/resources/usr/bin/validate_model.R
```diff
@@ -0,0 +1,431 @@
...
+                        "Missing factor levels for variable '", VARIABLE, "'. ",
+                        "Present levels: ", paste( names( expected_factors_levels[ expected_factors_levels ]), collapse = " " ), ". ",
+                        "Missing levels: ", paste( names( expected_factors_levels[!expected_factors_levels ]), collapse = " " )
+                        )
+                    )
+            }
+
+            ## Check whether there are levels in the table that are not present in the models definitions
+            if (!all( obtained_factors_levels[!is.na(obtained_factors_levels)] %in% names(expected_factors_levels)) ) {
+                warnings <- c(warnings,
```
suzannejin, line 304: are the warnings contrast id based? Just to trace in which contrast definition these warnings are happening

modules/local/validatemodel/resources/usr/bin/validate_model.R
```diff
@@ -0,0 +1,431 @@
...
+        # Gather blocking factors
+        if (!is.null(CONTRAST$blocking_factors)) {
+            blocking_vars <- unique(c(blocking_vars, CONTRAST$blocking_factors))
+        }
+    }
+
+    # Consolidate gathered contrasts into `var`
+    for (variable in names(temp_var)) {
+        if (variable %in% names(var)) {
+            var[[variable]] <- unique(c(var[[variable]], temp_var[[variable]]))
```
suzannejin: will this overwrite contrasts lists if there are two contrasts ids trying to contrast on the same variable, but different conditions? Eg. tissue 1 vs tissue 2, and then tissue 2 vs tissue 3?

modules/local/validatemodel/resources/usr/bin/validate_model.R
```diff
@@ -0,0 +1,431 @@
...
+                errors <- c(errors,
+                    paste0("Column ", VARIABLE, " contains undesired characters\n")
+                )
+            }
+
+            ## Extract the expected levels according to the YML file
+            expected_factors_levels <- c( rep(NA, length(variables[[ VARIABLE ]])) )
+
+            ## Assign names
+            names(expected_factors_levels) <- variables[[ VARIABLE ]]
```
suzannejin: so related to above question, it looks to me that this would overwrite the check if there are two contrasts for the same variable but different levels?

pinin4fjords (changes requested):
I'm afraid the local module won't work. Module bin has limitations on where it's functional (https://www.nextflow.io/docs/latest/module.html#module-binaries), which is why we don't use them.

The best way to approach R modules is still to use templates.

I normally discourage local modules, but if this is highly specific to this workflow it probably makes sense.

atrigila:
Thank you for your feedaback, @suzannejin @pinin4fjords. I'll talk to Alan who developed this module for some of these more specific questions about his design. I will also see how I can adapt this to the template.

modules/local/validatemodel/resources/usr/bin/validate_model.R
atrigila: It was all @alanmmobbs93 :) I also liked those

modules/local/validatemodel/resources/usr/bin/validate_model.R
atrigila: It is defined here:

https://github.com/atrigila/differentialabundance/blob/a5d0ea20167c69d17c72a786f4a8cda36411a418/modules/local/validatemodel/resources/usr/bin/validate_model.R#L102

modules/local/validatemodel/resources/usr/bin/validate_model.R
atrigila: In blocking factors you have column names only, while in the other variables you have a combination of variables plus levels. The content of each variable is different in blocking factors. 
CC @alanmmobbs93

modules/local/validatemodel/resources/usr/bin/validate_model.R
atrigila: No, that's why there is a `c()` to combine all:
`var[[variable]] <- unique(c(var[[variable]], temp_var[[variable]]))`
cc @alanmmobbs93

modules/local/validatemodel/resources/usr/bin/validate_model.R
alanmmobbs93: @suzannejin @atrigila Exactly that. If you check the yml structure, the arrays generated from "comparison" and "blocking factors" have different contents, that's why I treated them separately.
Heres an example yml used for testing [Test YML](https://raw.githubusercontent.com/nf-core/test-datasets/refs/heads/differentialabundance/testdata/MaxQuant_contrasts.yaml), just in case

modules/local/validatemodel/resources/usr/bin/validate_model.R
alanmmobbs93: @suzannejin At this point, the goal is to check that each level of a variable is present in the sample sheet. 

For example, lets assume the yml file presents a comparison of the variable `treatment`, with levels `control` vs `treatmentA`, and another comparison with levels `control` vs `treatmentB`. So, this block of code will check that the three `control`, `treatmentA`, and `treamentB` are present in the sample sheet column `treatment`. 
The three levels were collected into a list with the previous function `process_models()`. I think this is the part that should answer your concern, each variable is evaluated once with all its levels together.

Just let me know if I answered your question and thanks for taking time to look into this
CC @atrigila

atrigila:
Hi @suzannejin and @pinin4fjords , I have migrated the module to the template structure and answered the questions. Please feel free to take a look again if you can :)

modules/local/validatemodel/tests/nextflow.config
```diff
@@ -0,0 +1,5 @@
+process {
+    withName: 'VALIDATE_YML_MODEL' {
+        ext.args = params.module_args
```
pinin4fjords: Is there even a `module_args` parameter? Probably wouldn't make sense if there was.

modules/nf-core/custom/tabulartogseachip/environment.yml
```diff
@@ -1,7 +1,5 @@
----
```
pinin4fjords: Shouldn't the modules.json have changed? Did you forget to commit?

modules/local/validatemodel/templates/validate_model.R
```diff
@@ -0,0 +1,468 @@
+#!/usr/bin/env Rscript
+
+## TODO: add make contrast arguments
```
pinin4fjords: Remove TODOs

modules/local/validatemodel/templates/validate_model.R
```diff
@@ -0,0 +1,468 @@
...
+var            <- models_lists[[ "var" ]]
+
+## Print explicit messages
+for (INDEX in 1:length(var)) {
+    cat("\033[32mDetected '", names(var)[INDEX], "' variable with ", paste(var[[INDEX]], collapse = " "), " levels.\033[0m \n", sep = "")
+}
+
+# VALIDATE SAMPLESHEET BASED ON YML FILE ----------------------------
+## Define function to compare YML and samplesheet
+validate_model <- function(sample_column, variables, samplesheet) { # sample_column: string; variables: list; samplesheet: df
```
pinin4fjords: Can we move all functions to the top please? It's hard to read code that intersperses the functions and other code this way.

modules/local/validatemodel/resources/usr/bin/validate_model.R
suzannejin: thank you for explaining, it answered my doubt :)

modules/local/validatemodel/templates/validate_model.R
```diff
@@ -0,0 +1,468 @@
...
+
+## Process models
+models_lists   <- process_models(models)
+### Extract models contrasts
+contrasts_list <- models_lists[[ "contrasts_list" ]]
+### Extract variables and levels
+var            <- models_lists[[ "var" ]]
+
+## Print explicit messages
+for (INDEX in 1:length(var)) {
```
pinin4fjords: Again with the capitalised variable- that's not conventional, lowercase will be fine.

modules/local/validatemodel/templates/validate_model.R
pinin4fjords: Could we also have proper roxygen2 function docs please, rather than a non-standard comment style? Your favourite AI assistant can probably mostly do that for you. See other R modules for examples.

modules/local/validatemodel/templates/validate_model.R
```diff
@@ -0,0 +1,468 @@
+#!/usr/bin/env Rscript
+
```
pinin4fjords: Add a description about what the code is for, and what it does?

Ideally actually, we'd have a meta.yml. Realise it's not always done for local modules, but I think it would be useful

modules/local/validatemodel/templates/validate_model.R
```diff
@@ -0,0 +1,468 @@
...
+    # Add blocking variables to `var`
+    var[[ "blocking_factors" ]] <- blocking_vars
+
+    # Return the populated lists
+    return(list(
+        contrasts_list = contrasts_list,
+        var = var
+    ))
+}
```
pinin4fjords:
```suggestion
process_models <- function(models) {
  
  if (is.null(models$contrasts) || length(models$contrasts) == 0) {
    stop("models$contrasts is null or empty. Please provide valid input.")
  }
  
  contrasts_list <- list()
  temp_var <- list()
  blocking_vars <- c()
  
  for (contrast in models$contrasts) {
    
    if (is.null(contrast$id)) {
      stop("Missing contrast$id detected.")
    }
    
    if (contrast$id %in% names(contrasts_list)) {
      stop(sprintf("Duplicate contrast$id detected: %s.", contrast$id))
    }
    
    # Extract the variable name from the first element of the comparison vector.
    variable <- contrast$comparison[[1]]
    if (is.null(variable) || variable == "") {
      stop("Invalid variable defined in contrast$comparison[1].")
    }
    
    # Safely extract the remaining elements (if any) as levels.
    levels <- if (length(contrast$comparison) > 1) contrast$comparison[-1] else NULL
    
    contrasts_list[[contrast$id]] <- list(
      variable = variable,
      contrast = levels,
      blocking_factors = contrast$blocking_factors
    )
    
    # Append and deduplicate levels for the variable.
    if (!is.null(temp_var[[variable]])) {
      temp_var[[variable]] <- unique(c(temp_var[[variable]], levels))
    } else {
      temp_var[[variable]] <- levels
    }
    
    # Accumulate blocking factors.
    if (!is.null(contrast$blocking_factors)) {
      blocking_vars <- unique(c(blocking_vars, contrast$blocking_factors))
    }
  }
  
  # Use the temporary variable contrasts and add blocking factors.
  var <- temp_var
  var[["blocking_factors"]] <- blocking_vars
  
  list(
    contrasts_list = contrasts_list,
    var = var
  )
}
```

pinin4fjords: AI-proposed simplification, bit more concise

tests/test_soft.nf.test.snap
```diff
@@ -76,15 +87,19 @@
                 "app.R:md5,bedcfc45b6cdcc2b8fe3627987e2b17a",
                 "versions.yml:md5,27cf350d1f743e60e64b4434cf6d4687",
                 "normalised.annotation.tsv:md5,40300ae222ae35fa54daf10a1b86e830",
-                "phenotype_uninvolved_lesional_study.limma.results.tsv:md5,2cfa653fffb034be4b82abf306c87bc5",
-                "phenotype_uninvolved_lesional_study.limma.results_filtered.tsv:md5,6a198c35b193bab0b3148d147c7b08e3",
-                "normalised.matrix.tsv:md5,230f87f91c0c1a7e2ab354200c4b978c"
+                "phenotype_uninvolved_lesional.limma.results.tsv:md5,2cfa653fffb034be4b82abf306c87bc5",
```
pinin4fjords: Why have some output file names changed?

pinin4fjords (changes requested):
Made a few comments. General points:

- We need to make sure output files and md5sums are not changed unless there's a good reason. 
- The R code is a bit scrappy, we should try and make it nicer and more idiomatic. I've attached an AI-assisted version here which I think is closer to the mark and cleaner, it might need some final debugging. 

[foo.r.zip](https://github.com/user-attachments/files/18941281/foo.r.zip)

modules/local/validatemodel/templates/validate_model.R
pinin4fjords: Actually ignore this- I remade the whole script, look there instead

atrigila: I agree. I used your provided `.zip` refactored code and it worked.

modules/local/validatemodel/templates/validate_model.R
atrigila: I agree. Added a `meta.yml`.

modules/local/validatemodel/templates/validate_model.R
atrigila: I agree. I used your provided `.zip` refactored code and it worked.

modules/local/validatemodel/templates/validate_model.R
atrigila: Solved in the new refactored code.

modules/local/validatemodel/templates/validate_model.R
atrigila: Solved in the new refactored code.

modules/local/validatemodel/templates/validate_model.R
atrigila: Solved in the new refactored code.

modules/nf-core/custom/tabulartogseachip/environment.yml
atrigila: This was a change to `environment.yml` only due to linting issues I believe. I reverted the file so that it is equal to the module.

modules/local/validatemodel/tests/nextflow.config
atrigila: Changed the test params so that they more accurately reflect the used params in the pipeline.

tests/test_soft.nf.test.snap
atrigila: This was a temporary change after merging the dev branch from the GitHub editor. The current tests do not change output names and do not differ to previous tests from dev (except ofc the new yaml tests that were added).

modules/local/validatemodel/templates/validate_model.R
```diff
@@ -0,0 +1,420 @@
...
+        paste('    tidyverse:', tidyverse_version)
+    ),
+    'versions.yml'
+)
+
+################################################
+################################################
+## SOFTWARE VERSIONS                           ##
+################################################
+################################################
```
pinin4fjords:
```suggestion
```

My bad

workflows/differentialabundance.nf
```diff
@@ -193,6 +195,15 @@ workflow DIFFERENTIALABUNDANCE {
             }
             .flatten()
             .unique() // Uniquify to keep each contrast variable only once (in case it exists in multiple lines for blocking etc.)
+
+        VALIDATE_YML_MODEL (
```
pinin4fjords: Sorry, one last thought. 

The issue with doing it this way is that `ch_input` etc will proceed through other processes even if there are problems that eventually get flagged by this process. What we really want is for this to stop things before they get that far.

You can use a dummy join to make that happen- I do that in [this subworkflow](https://github.com/nf-core/modules/blob/4d95826db394fc94c9e9ea11170bbe19e0d849ca/subworkflows/nf-core/fastq_qc_trim_filter_setstrandedness/main.nf#L156) for example. Or you can just have VALIDATE_YML_MODEL spit the matrix and contrasts back out again.

pinin4fjords (changes requested):
Another comment, but in making it I realised I'd already thought about it when making the VALIDATEFOMCOMPONENTS script. 

Sorry to be slow on thinking of this (other things going on), but is there a reason you didn't simply extend that other validation process? Seems confusing for the user/ maintainer to have two validation steps. 

Work was already done to enable yaml contrasts there: https://github.com/pinin4fjords/shinyngs/pull/68

[This script](https://github.com/pinin4fjords/shinyngs/blob/develop/exec/validate_fom_components.R) uses validation functions built into the parsing of the various objects themselves, and that will probably be a neater way to go rather than duplicating validation logic here.

Sorry to be a pain, I can help with this when I get some time and we can bake the work you've done here there. We probably just need to extend the existing logic that validates contrasts here: https://github.com/pinin4fjords/shinyngs/blob/931d9c39c1f2200c66cc628e1ea8d68e970262ef/R/accessory.R#L908

workflows/differentialabundance.nf
pinin4fjords: Work was already done to enable yaml contrasts there: https://github.com/pinin4fjords/shinyngs/pull/68

nschcolnicov:
> Another comment, but in making it I realised I'd already thought about it when making the VALIDATEFOMCOMPONENTS script.
> 
> Sorry to be slow on thinking of this (other things going on), but is there a reason you didn't simply extend that other validation process? Seems confusing for the user/ maintainer to have two validation steps.
> 
> Work was already done to enable yaml contrasts there: [pinin4fjords/shinyngs#68](https://github.com/pinin4fjords/shinyngs/pull/68)
> 
> [This script](https://github.com/pinin4fjords/shinyngs/blob/develop/exec/validate_fom_components.R) uses validation functions built into the parsing of the various objects themselves, and that will probably be a neater way to go rather than duplicating validation logic here.
> 
> Sorry to be a pain, I can help with this when I get some time and we can bake the work you've done here there. We probably just need to extend the existing logic that validates contrasts here: https://github.com/pinin4fjords/shinyngs/blob/931d9c39c1f2200c66cc628e1ea8d68e970262ef/R/accessory.R#L908

@pinin4fjords I can answer that. We decided to create a new script instead of updating the shinyngs validatefromcomponents module because maintaining a separate tool just to run an R script adds unnecessary complexity and significantly slows down development. Since this process can be handled effectively with a standalone R script, it makes more sense to simplify our workflow rather than having to update and maintain an entire tool every time we need to make changes. Using an R script and module streamlines development, and if we ever need more robustness, we have the option to integrate it into nf-core.

pinin4fjords:
> > Another comment, but in making it I realised I'd already thought about it when making the VALIDATEFOMCOMPONENTS script.
> > Sorry to be slow on thinking of this (other things going on), but is there a reason you didn't simply extend that other validation process? Seems confusing for the user/ maintainer to have two validation steps.
> > Work was already done to enable yaml contrasts there: [pinin4fjords/shinyngs#68](https://github.com/pinin4fjords/shinyngs/pull/68)
> > [This script](https://github.com/pinin4fjords/shinyngs/blob/develop/exec/validate_fom_components.R) uses validation functions built into the parsing of the various objects themselves, and that will probably be a neater way to go rather than duplicating validation logic here.
> > Sorry to be a pain, I can help with this when I get some time and we can bake the work you've done here there. We probably just need to extend the existing logic that validates contrasts here: https://github.com/pinin4fjords/shinyngs/blob/931d9c39c1f2200c66cc628e1ea8d68e970262ef/R/accessory.R#L908
> 
> @pinin4fjords I can answer that. We decided to create a new script instead of updating the shinyngs validatefromcomponents module because maintaining a separate tool just to run an R script adds unnecessary complexity and significantly slows down development. Since this process can be handled effectively with a standalone R script, it makes more sense to simplify our workflow rather than having to update and maintain an entire tool every time we need to make changes. Using an R script and module streamlines development, and if we ever need more robustness, we have the option to integrate it into nf-core.

OK, but I disagree with that call. 

I'm not asking for the creation of a new script within shinyngs. The fact is, validation functionality is already present, the nf-core module already exists. All this is is some additional logic for that existing validation function. It doesn't make sense to create a new module partially replicating that logic and adding some new. 

This code needs to go into the existing Shinyngs function, as stated above. That function already has access to the sample sheet and contrasts, so I don't see that being all that difficult. I will assist on PRs and releases to make that happen.

grst:
I'm wondering if it could be a provisional solution to have it as a separate script in the differentialabundance pipeline? We are planning several more interations on this in #429, #377 and #386 and going through shinyngs releases for each step is really really cumbersome. I'm all for putting things in their proper place in the end, but this process is really slowing down development for questionable benefit.

pinin4fjords:
> I'm wondering if it could be a provisional solution to have it as a separate script in the differentialabundance pipeline? We are planning several more interations on this in #429, #377 and #386 and going through shinyngs releases for each step is really really cumbersome. I'm all for putting things in their proper place in the end, but this process is really slowing down development for questionable benefit.

The benefits aren't questionable for me, they're tangible and I did this for a reason. This was a design decision I made when finalising the first versions of this pipeline, and we transitioned from a development to a production mindset. It allows us to share e.g. parsing logic between processes, and keeps scientific logic out of the workflow, which is then just doing orchestration. We also try and avoid local components, as you will have noted, and this facilitates that. 

I'm sorry, but I'm not prepared to reverse those benefits just to shorten the development loop, this sort of overhead is not atypical when doing further development on tools in a production state. 

But as I say I will help with the shinyngs PRs and releases, as I've done previously.

pinin4fjords:
Modules updated in https://github.com/nf-core/modules/pull/7523. This PR can be changed to just be a module bump in the workflow to pull that (once it's merged).

pinin4fjords (approved):
Great, thanks for sorting this clean solution

PR #444, opened 2025-03-17
Fix: Quote shell parameters

zeehio:
I ran a differential abundance pipeline with a study name that had spaces in it, and the pipeline crashed because the `make_app_from_files.R` caller script was not escaping parameters as expected. I believe the patch included in this PR is pretty easy to follow: The parameters were expanded by the `$` and the second word of the study name was interpreted as another argument.

Making this change made the pipeline succeed.

If you believe this contribution is of your interest but you require of me to add tests I can try to find time to learn how to write tests and comply with the rest of the checklist. It may take me a while though to find the time, and the patch is really minor so I would appreciate your consideration.

You may want to consider security ramifications of being able to pass a study_name like `; arbitrary_command` and inject code that would be executed by the pipeline. I am not familiar enough with nextflow to understand whether that should be considered a vulnerability risk or not, but if you think this is an issue, then this PR would fix it.

In any case, thanks for your time and for all the work you do. 

Looking forward to getting more involved!

## PR checklist

### DIFF
--- a/modules/nf-core/shinyngs/app/main.nf
+++ b/modules/nf-core/shinyngs/app/main.nf
@@ -38,13 +38,13 @@ process SHINYNGS_APP {
 
     """
     make_app_from_files.R \\
-        --sample_metadata $sample \\
-        --feature_metadata $feature_meta \\
-        --assay_files ${assay_files.join(',')} \\
-        --contrast_file $contrasts \\
-        --contrast_stats_assay $contrast_stats_assay \\
-        --differential_results ${differential_results.join(',')} \\
-        --output_dir $prefix \\
+        --sample_metadata "$sample" \\
+        --feature_metadata "$feature_meta" \\
+        --assay_files "${assay_files.join(',')}" \\
+        --contrast_file "$contrasts" \\
+        --contrast_stats_assay "$contrast_stats_assay" \\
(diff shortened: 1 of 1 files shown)


pinin4fjords:
The source for this script is at https://github.com/pinin4fjords/shinyngs/blob/develop/exec/make_app_from_files.R. Could you make the PR there please? Then you'll get proper credit.

zeehio:
Hi @pinin4fjords,

My PR does not touch the R script, which I believe it is fine as it is, but rather to the nextflow script that calls it.

That's why I believe the patch belongs here

pinin4fjords:
@zeehio sorry, yes, you're (partly) right, I read too quickly.

But the module will need to be updated in nf-core/modules, then we'll update it here.

zeehio:
> @zeehio sorry, yes, you're (partly) right, I read too quickly.
> 
> But the module will need to be updated in nf-core/modules, then we'll update it here.

Aha, is this it?

https://github.com/nf-core/modules/pull/8177

pinin4fjords:
Yep! Now you just have to appease your reviewer :-)

grst:
Module is now merged, this requires a simple module update to get this in now.

atrigila:
Closing as addressed in #653

PR #443, opened 2025-03-06
Add toolsheet-related implementations

suzannejin:
Changes made to enable the usage of toolsheets:
- [x] Add predefined toolsheets -> they define the combinations of tools allowed for each study type
- [x] Add toolsheet_custom flag -> needed later for benchmark functionalities
- [x] Modify `gsea_run` and `gprofiler2_run` flags into `functional_method` -> more consistent with the ch_tools behaviour
- [x] Update report rmd file to handle `differential_method` and `functional_method` flags
- [x] Add code to parse ch_tools from toolsheets
- [x] Handle the results channels inside the workflow in a way that it considers the tools used
- [x] Make ch_tools args accesible by modules.config through meta.args

Nice to have, but left for next PRs (to not extend this one too much):
- [ ] Replace the checks on params.functional_method, etc by checks on ch_tools
- [ ] Add checks for `analysis_name` flag
- [ ] Remove tool-specific fixed params (eg. differential_file_suffix, etc) from user params scope (having too many params is confusing) and have them defined inside the workflow based on the chosen method.
- [ ] Add tests for gprofiler2 runs
- [ ] Clean some redundant checks/code/tests related to study_type vs tool combination - before it must be checked, but now we have the predefined tools that dictate which combinations can go together.
- [ ] Create ch_background based on ch_tools instead of params

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core lint`).
- [x] Ensure the test suite passes (`nf-test test main.nf.test -profile test,docker`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [x] `README.md` is updated (including new tool citations and authors/contributors).

PR #446, opened 2025-03-26
Add `variancepartition/dream`

atrigila:
https://github.com/nf-core/differentialabundance/issues/411

## PR checklist

### DIFF
--- a/conf/rnaseq_dream.config
+++ b/conf/rnaseq_dream.config
@@ -31,6 +31,5 @@ params {
     differential_qval_column         = "adj.P.Val"
     differential_feature_id_column   = "Geneid"
     differential_feature_name_column = "Geneid"
-    limma_use_voom                   = true
 
 }
--- a/conf/rnaseq_dream.config
+++ b/conf/rnaseq_dream.config
@@ -26,8 +26,8 @@ params {
     // Differential options
     differential_use_dream           = true
     differential_file_suffix         = ".dream.results.tsv"
-    differential_fc_column           = "logFC"                  // TODO: To be discussed
-    differential_pval_column         = "P.Value"                // TODO: To be discussed
+    differential_fc_column           = "logFC"
+    differential_pval_column         = "P.Value"
     differential_qval_column         = "adj.P.Val"
     differential_feature_id_column   = "Geneid"
     differential_feature_name_column = "Geneid"
--- a/conf/rnaseq_dream.config
+++ b/conf/rnaseq_dream.config
@@ -1,0 +1,36 @@
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Nextflow config file for running RNA-seq analysis
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    Defines settings specific to RNA-seq analysis
+
+    Use as follows:
+        nextflow run nf-core/differentialabundance -profile rnaseq_dream,<docker/singularity> --outdir <OUTDIR>
+
+----------------------------------------------------------------------------------------
+*/
+
+params {
+
+    config_profile_name        = 'RNA-seq profile with `VARIANCEPARTITION_DREAM`'
+    config_profile_description = 'Settings for RNA-seq analysis with `VARIANCEPARTITION_DREAM`'
+
+    // Study
+    study_type = 'rnaseq'
+    study_abundance_type = 'counts'
+
+    // Observations
+    observations_id_col        = 'sampleId'
+    observations_name_col      = 'sampleId'
+
+    // Differential options
+    differential_use_dream           = true
+    differential_file_suffix         = ".dream.results.tsv"
+    differential_fc_column           = "logFC"                  // TODO: To be discussed
+    differential_pval_column         = "P.Value"                // TODO: To be discussed
+    differential_qval_column         = "adj.P.Val"
+    differential_feature_id_column   = "Geneid"
+    differential_feature_name_column = "Geneid"
+    limma_use_voom                   = true
+
+}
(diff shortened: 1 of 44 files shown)


subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -179,11 +189,20 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
         ch_diff_filter_params.fc_input,
         ch_diff_filter_params.stat_input
     )
+    ch_versions = ch_versions.mix(CUSTOM_FILTERDIFFERENTIALTABLE.out.versions.first())
+
+    ch_dream_results = Channel.empty()
+
+    VARIANCEPARTITION_DREAM(ch_contrasts_dream)
+    ch_versions = ch_versions.mix( VARIANCEPARTITION_DREAM.out.versions.first() )
+    ch_dream_results = VARIANCEPARTITION_DREAM.out.results
```
pinin4fjords: What is the reason for new new method to have its own output channel, when all other methods direct their results to the same `ch_results` channel:

pinin4fjords (changes requested):

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -26,9 +27,12 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
     ch_transcript_lengths    // [ meta_exp, transcript_lengths]
     ch_control_features      // [meta_exp, control_features]
     ch_contrasts             // [[ meta_contrast, contrast_variable, reference, target ]]
+    ch_contrasts_dream       // [meta_contrast, samplesheet, matrix]
```
pinin4fjords: In fact, why does it have its own input as well? By doing this, you exclude DREAM from the combinatorial ways of setting things up that come after, which are central to the functionality of the subworkflow.

pinin4fjords:
The subworkflow definitely needs to be fixed to remove the custom channels, or it breaks the iterative stuff @suzannejin has been working on.

subworkflows/nf-core/abundance_differential_filter/main.nf
atrigila: I was transferring the work from the previous PR into this one, and while this point wasn’t raised earlier during its review, I agree with your observation. I’m not aware of the reason behind this decision.

My next steps are to remove this from the subworkflow. If I find a blocker that justifies having it as its own channel, I’ll update here.

grst: I believe it's simply that the new structure hadn't been adopted yet when the tmp_dev branch was developed...

atrigila: I did some investigation into why `ch_contrasts` might not have been used, and I believe the issue stems from how the YAML contrasts file is handled in the `VALIDATOR` module. Specifically, the formula field is not emitted as part of the `VALIDATOR.out.contrasts` and is therefore lost. 

I propose the following to move forwards:
-  Modify `shinyngs/validatefomcomponents` R script to emit a new column, formula, in YAML contrast files. The output in `VALIDATOR.out.contrasts` will then be:
```
id	variable	reference	target	blocking	formula
treatment_mCherry_hND6_sample_number	treatment	mCherry	hND6	sample_number	~ treatment + sample_number
```
We'll then have `meta.formula` in `ch_contrasts` (and I'll reassess how to continue from there).

atrigila: Quick update: I’ve [opened a PR to shinyngs](https://github.com/pinin4fjords/shinyngs/pull/72) to address the missing formula field issue. I’ve also already started making some changes in this draft adjusting the subworkflows and preparing for integration.

Once the `shinyngs` update is merged, I’ll test it here, complete the necessary subworkflow modifications, and add a new test case. If everything works as expected and there’s agreement on the changes, I’ll move on to updating the corresponding subworkflows upstream in nf-core/modules.

conf/rnaseq_dream.config
```diff
@@ -0,0 +1,35 @@
...
+    // Study
+    study_type = 'rnaseq'
+    study_abundance_type = 'counts'
+
+    // Observations
+    observations_id_col        = 'sampleId'
+    observations_name_col      = 'sampleId'
+
+    // Differential options
+    differential_use_dream           = true
```
pinin4fjords, line 27: This is OK for now. I think once we get @suzannejin 's work in , this will be simplified like this https://github.com/nf-core/differentialabundance/blob/61dea79f900eb7f0162feb070f5f32eaa833df28/nextflow.config#L98 (see #443)

workflows/differentialabundance.nf
```diff
@@ -92,9 +92,9 @@ if (params.study_type == 'rnaseq') {
 // Use 'limma' for specific study types or RNAseq with limma flag; else 'deseq2'
 // Also set fold change and q-value thresholds.
 tools_differential = [
-    method        : ((params.study_type in ['affy_array', 'geo_soft_file', 'maxquant']) ||
-                    (params.study_type == 'rnaseq' && params.differential_use_limma))
-                    ? 'limma' : 'deseq2',
+    method: (params.study_type == 'rnaseq' && params.differential_use_dream) ? 'dream' :
```
pinin4fjords, line 95: Again, this logic gets out of hand as we add new methods, we can shift to a simpler approach in the near future, but this is consistent with the current logic.

conf/rnaseq_dream.config
```diff
@@ -0,0 +1,35 @@
...
+    differential_file_suffix         = ".dream.results.tsv"
+    differential_fc_column           = "logFC"                  // TODO: To be discussed
```
pinin4fjords: Let's discuss then :-)

conf/rnaseq_dream.config
atrigila: Makes a lot of sense! I was actually wondering if there was a reason this pipeline didn't have a `--tools` argument or similar, but glad to see this is coming :)

conf/rnaseq_dream.config
atrigila: Sorry, forgot to remove this. I think I solved it along the way.

pinin4fjords:
Still marked as draft, are you still working?

pinin4fjords (approved):
LGTM

PR #448, opened 2025-03-31
NOMERGE simplify toolsheethandling

### LINKED ISSUE #423: Use toolsheet channel to handle the calls to methods

pinin4fjords:
My reworking of https://github.com/nf-core/differentialabundance/pull/443, for illustration in PR review there. 

Note: **I've stripped out a lot of workflow code, and am progressively re-adding to get to the state of #443, but with the new formulation **

## PR checklist


subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -135,6 +142,63 @@ workflow PIPELINE_COMPLETION {
...
+        error("Input samplesheet not specified!")
+    }
+
+    // Validate study type specific parameters
+    if (params.study_type == 'affy_array') {
+        if (!params.affy_cel_files_archive) {
+            error("CEL files archive not specified!")
+        }
+    } else if (params.study_type == 'maxquant') {
+        if (params.functional_method) {
```
grst, line 197: Also not directly related: I don't know a lot about Proteomics data, but why would it not be possible to run a functional method? In the end it produces also a matrix with expression values, isn't it?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -135,6 +142,63 @@ workflow PIPELINE_COMPLETION {
...
+        }
+    }
+
+    // Check mandatory parameters
+    if (!params.input) {
+        error("Input samplesheet not specified!")
+    }
+
+    // Validate study type specific parameters
+    if (params.study_type == 'affy_array') {
```
grst, line 192: Not directly related to this PR, but I haven't fully understood the scope of `study_type`. Is it only to specify which data input type to use? Or is it also to define which methods to run on which data? If the latter, wouldn't that be the point of the new toolsheet? 

I'm mostly asking because I think I'd have to choose "rnaseq" for running limma/dream on an input matrix, that doesn't necessarily need to be rnaseq/count data.

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -240,167 +304,123 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
-    if (params.analysis_name) {
-        return getToolsheetConfigurations()
-    } else {
-        return getDefaultConfigurations()
+        if (!analysis_name_def) {
+            error("Could not find analysis_name definition in pipeline schema")
+        }
+
+        // Add analysis_name to properties with meta field
+        all_properties['analysis_name'] = analysis_name_def + [meta: ['analysis_name']]
```
grst: Why is this part necessary? I understood you already fail if the toolsheet header doesn't contain `analysis_name` and copy over all properties that are in the header from the pipeline_schema into `all_properties`. 

Couldn't you just fail if `analysis_name` is not in `all_properties`?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -240,167 +304,123 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
-    def args = parseParams('differential', method, tool_config?.diff_args ?: null)
-    def fc_threshold = args.differential_min_fold_change
-    def stat_threshold = args.differential_max_qval
-    return [
-        method: method,
-        args: args,
-        fc_threshold: fc_threshold,
-        stat_threshold: stat_threshold
-    ]
+    return toolsheet.collect{ row -> (params.findAll { k, v -> k != 'trace_report_suffix' } + row[0]) }
```
grst: I don't follow what this does, consider adding a comment.

grst (commented):
Hi @pinin4fjords and @suzannejin, 

I like the direction this is going! 

>  I propose we simplify this by making the toolsheet directly represent pipeline parameters. 
> Each row represents a complete set of pipeline parameters

This sounds a whole lot like nextflow profiles. With the difference being that multiple "profiles" can be run in a single pipeline invokation? 

I'm also still trying to understand how the toolsheet would look like for my use-cases. If I want to run RNA-seq analysis with limma/voom + apply different functional analysis tools (e.g. gsea on gene ontology + decoupler on 3 different signatures), would that look somewhat like

```csv
analysis_name,study_type,differential_method,limma_use_voom,functional_method,signature_file
limma_rnaseq,rnaseq,limma,true,gsea,gene_ontology.gmt
limma_rnaseq,rnaseq,limma,true,decoupler,progeny.tsv
limma_rnaseq,rnaseq,limma,true,decoupler,collectri.tsv
limma_rnaseq,rnaseq,limma,true,decoupler,hallmark.tsv
```

?


Would that then run limma four times with the same settings, or run it once and reuse the results for the different functional analyses?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: That's a very good point. We could reduce the usage of `study_type` in some contexts, where a parameter to drive a particular decision point might be more appropriate.

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: We can definitely look at tidying up this function- consider this a first go to demonstrate the principle.

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: Think of `collect` like a `map` in Nextflow. `trace_report_suffix` is generated at runtime and it was invalidating the caches every time, so it's import to not include it. Will add comment.

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: @WackerO established this logic. Proteomics data is a bit weird in its sparsity etc, maybe the assumptions associated with GSEA etc don't hold.

pinin4fjords:
> This sounds a whole lot like nextflow profiles. With the difference being that multiple "profiles" can be run in a single pipeline invokation?

Yep, pretty much.

> I'm also still trying to understand how the toolsheet would look like for my use-cases. If I want to run RNA-seq analysis with limma/voom + apply different functional analysis tools (e.g. gsea on gene ontology + decoupler on 3 different signatures), would that look somewhat like
> 
> ```
> analysis_name,study_type,differential_method,limma_use_voom,functional_method,signature_file
> limma_rnaseq,rnaseq,limma,true,gsea,gene_ontology.gmt
> limma_rnaseq,rnaseq,limma,true,decoupler,progeny.tsv
> limma_rnaseq,rnaseq,limma,true,decoupler,collectri.tsv
> limma_rnaseq,rnaseq,limma,true,decoupler,hallmark.tsv
> ```
> 
> ?
> 
> Would that then run limma four times with the same settings, or run it once and reuse the results for the different functional analyses?

As coded in this first pass, it would run Limma 4x, because the `meta` for the process will be different. Once solution I'm thinking of will be to use subsets of the metadata to collapse inputs for certain steps (using a groupTuple), and then re-expand afterwards (using a `join()` and `transpose()`) so we maintain the association between results and paramsets without duplicating work. That will require us to maintain a dictionary of the params pertinent to the various steps. I'll add something to illustrate that today.

This is related to what @suzannejin is proposing in separating differential, functional params etc, but I think we can implement it more cleanly in this setup.

grst:
> As coded in this first pass, it would run Limma 4x, because the meta for the process will be different. Once solution I'm thinking of will be to use subsets of the metadata to collapse inputs for certain steps (using a groupTuple), and then re-expand afterwards (using a join() and transpose()) so we maintain the association between results and paramsets without duplicating work. That will require us to maintain a dictionary of the params pertinent to the various steps. I'll add something to illustrate that today.

One other approach could be to not fundamentally distinguish between differential and functional methods, i.e. both are just methods and use an input value which can either be the input matrix, or the output of another matrix. This would also accomodate that certain functional methods like decoupler can be executed both, on the gene expression matrix, and the differential expression results. 


```bash
analysis_name,method,source,limma_use_voom,signature_file
# run limma on gene expression matrix
limma_rnaseq,limma,GEX_MATRIX,true,
# run decoupler on limma result. `source` references another "analysis name"
decoupler_rnaseq_de_progeny,decoupler,limma_rnaseq,,progeny.tsv
decoupler_rnaseq_de_collectri,decoupler,limma_rnaseq,,collectri.tsv
# run decoupler on gene expression matrix
decoupler_rnaseq_progeny,decoupler,GEX_MATRIX,,progeny.tsv
# run limma on decoupler results (if decoupler is run on GEX matrix, it generates separate values for each sample, we still need to do stats)
limma_progeny,limma,decoupler_rnaseq_progeny,false,
```

pinin4fjords:
> > As coded in this first pass, it would run Limma 4x, because the meta for the process will be different. Once solution I'm thinking of will be to use subsets of the metadata to collapse inputs for certain steps (using a groupTuple), and then re-expand afterwards (using a join() and transpose()) so we maintain the association between results and paramsets without duplicating work. That will require us to maintain a dictionary of the params pertinent to the various steps. I'll add something to illustrate that today.
> 
> One other approach could be to not fundamentally distinguish between differential and functional methods, i.e. both are just methods and use an input value which can either be the input matrix, or the output of another matrix. This would also accomodate that certain functional methods like decoupler can be executed both, on the gene expression matrix, and the differential expression results.
> 
> ```shell
> analysis_name,method,source,limma_use_voom,signature_file
> # run limma on gene expression matrix
> limma_rnaseq,limma,GEX_MATRIX,true,
> # run decoupler on limma result. `source` references another "analysis name"
> decoupler_rnaseq_de_progeny,decoupler,limma_rnaseq,,progeny.tsv
> decoupler_rnaseq_de_collectri,decoupler,limma_rnaseq,,collectri.tsv
> # run decoupler on gene expression matrix
> decoupler_rnaseq_progeny,decoupler,GEX_MATRIX,,progeny.tsv
> # run limma on decoupler results (if decoupler is run on GEX matrix, it generates separate values for each sample, we still need to do stats)
> limma_progeny,limma,decoupler_rnaseq_progeny,false,
> ```

That's moving us back to a 'special thing' again, rather than a paramter set, which is Suzanne's solution, and adds a whole bunch of overheads, complex joins etc, which I really don't like.

By constraining this to be a simple parameter set, it's easy to understand and we open up a whole bunch of other use cases. We just have to mitigate the impact on repeated operations.

grst:
> That's moving us back to a 'special thing' again, rather than a paramter set

not necessarily, it just depends on how you define the parameters. E.g. the "source" would correspond to the `--matrix` flag with the only difference that it can reference another `analysis_name` in addition to a TSV/CSV file. Everything else could be standard parameters as fas as i can tell. 

> and adds a whole bunch of overheads, complex joins etc, which I really don't like.

this is probably the case, I haven't really thought it through how bad it would be.

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -329,29 +247,123 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
+
+    if (toolsheet.isEmpty()) {
+        if (params.analysis_name) {
+            error("No configuration found in toolsheet for analysis_name '${params.analysis_name}'")
+        } else {
+            error("No valid configurations found in toolsheet")
+        }
+    }
+
+    return toolsheet.map{it + params}
```
suzannejin: maybe params + it insted, so that toolsheet params stay in the highest priority?

pinin4fjords: Yep, I spotted that too- corrected now!

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -240,167 +304,123 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
+// 2. Extracting only those properties from the pipeline schema
+// 3. Creating a schema that allows for an array of objects
+// 4. Making analysis_name the only required field
+// 5. Adding meta fields to all properties
+// @return The absolute path to the temporary schema file
+def createToolsheetSchema() {
+    // Load toolsheet and get headers
+    def toolsheet_path = params.toolsheet_custom ?: "${projectDir}/assets/toolsheet.csv"
+    def toolsheet_lines = new File(toolsheet_path).readLines()
+    def headers = toolsheet_lines[0].split(',')
```
suzannejin, line 410: maybe also consider csv/tsv?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
suzannejin: i don't understand why collect and add row[0]?

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -25,14 +29,14 @@ workflow ABUNDANCE_DIFFERENTIAL_FILTER {
     ch_samplesheet           // [ meta_exp, samplesheet ]
     ch_transcript_lengths    // [ meta_exp, transcript_lengths]
     ch_control_features      // [meta_exp, control_features]
-    ch_contrasts             // [[ meta_contrast, contrast_variable, reference, target ]]
+    ch_contrasts             // [[ [meta_contrast], [contrast_variable], [reference], [target] ]]
```
suzannejin, line 32: why change it to this list structure?

pinin4fjords: This is because I wanted to allow for the possibility of a _different_ set of contrasts on each row of the toolsheet. I realise it's not your use case, but it is something this setup allows for. 

To do that, the contrasts have to be passed as one unit, in this structure. Then there's a quick `transpose()` you'll see further down to expand the contrasts again.

grst:
>> and adds a whole bunch of overheads, complex joins etc, which I really don't like.
> 
> this is probably the case, I haven't really thought it through how bad it would be.

Given the *constraint that only two methods can be chained*, I think it would not be that bad. 

(Very simplified) Pseudocode
```
toolsheet_step1 = ch_toolsheet.filter( it['input'] == "INPUT_MATRIX" }
toolsheet_step2 = ch_toolsheet.filter( it['input'] != "INPUT_MATRIX }

result_step1 = apply_tools(toolsheet_step1)
result_step2 = apply_tools(toolsheet_step2.join(result_step1, by={ x, y => x["tool_name"] == y["input"] }))
```

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -240,167 +304,123 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
-    def input_type = method == 'gprofiler2' ? 'filtered' :
-                    method == 'gsea' ? 'norm' :
-                    null
-    return [
-        method: method,
-        args: args,
-        input_type: input_type
-    ]
+// Get default configurations from pipeline parameters
+def getDefaultConfigurations() {
```
suzannejin, line 424: in my previous implementation i wanted to only parse the params that are relevant for the differential/functional analysis blocks, meaning the parameters with pattern 'differential|functional|deseq2|limma|gsea|etc". But it seems to me that you are aiming for parsing the entire params scope into ch_paramset?

workflows/differentialabundance.nf
```diff
@@ -187,182 +153,221 @@ workflow DIFFERENTIALABUNDANCE {
...
 
-        // Re-map the proteus output tables to the study ID as the tables are the same across contrasts, only one norm table will be necessary
-        ch_in_raw = PROTEUS.out.raw_tab
-            .first()
-            .map{ meta, matrix -> tuple(exp_meta, matrix) }
-        ch_in_norm = PROTEUS.out.norm_tab
-            .first()
-            .map{ meta, matrix -> tuple(exp_meta, matrix) }
+    ch_in_raw = ch_input.rnaseq
+        .map{meta, file -> [meta, meta.matrix]}
```
suzannejin, line 203: this is not clear to me why you take meta.matrix (the path str?) instead of the file?

suzannejin:
> > This sounds a whole lot like nextflow profiles. With the difference being that multiple "profiles" can be run in a single pipeline invokation?
> 
> Yep, pretty much.
> 
> > I'm also still trying to understand how the toolsheet would look like for my use-cases. If I want to run RNA-seq analysis with limma/voom + apply different functional analysis tools (e.g. gsea on gene ontology + decoupler on 3 different signatures), would that look somewhat like
> > ```
> > analysis_name,study_type,differential_method,limma_use_voom,functional_method,signature_file
> > limma_rnaseq,rnaseq,limma,true,gsea,gene_ontology.gmt
> > limma_rnaseq,rnaseq,limma,true,decoupler,progeny.tsv
> > limma_rnaseq,rnaseq,limma,true,decoupler,collectri.tsv
> > limma_rnaseq,rnaseq,limma,true,decoupler,hallmark.tsv
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> > Would that then run limma four times with the same settings, or run it once and reuse the results for the different functional analyses?
> 
> As coded in this first pass, it would run Limma 4x, because the `meta` for the process will be different. Once solution I'm thinking of will be to use subsets of the metadata to collapse inputs for certain steps (using a groupTuple), and then re-expand afterwards (using a `join()` and `transpose()`) so we maintain the association between results and paramsets without duplicating work. That will require us to maintain a dictionary of the params pertinent to the various steps. I'll add something to illustrate that today.
> 
> This is related to what @suzannejin is proposing in separating differential, functional params etc, but I think we can implement it more cleanly in this setup.

Exactly. One can achieve that by playing a bit with the joining/extension of channels based on the common paramset.

A downside, however, is that when a distinguishable row id (ie. analysis_name) is present inside the meta (and hence having different meta will run limma 4x), one can always save the output based on that id, otherwise the alternative is to save the output based on the row paramset, which can become not very readable for the user. Although I guess this can also be solved by not saving with publishdir at each module level, but use a final gather module to gather all the files generated for a given paramset row, and save the files at the end using the analysis_name as id.

suzannejin:
> > As coded in this first pass, it would run Limma 4x, because the meta for the process will be different. Once solution I'm thinking of will be to use subsets of the metadata to collapse inputs for certain steps (using a groupTuple), and then re-expand afterwards (using a join() and transpose()) so we maintain the association between results and paramsets without duplicating work. That will require us to maintain a dictionary of the params pertinent to the various steps. I'll add something to illustrate that today.
> 
> One other approach could be to not fundamentally distinguish between differential and functional methods, i.e. both are just methods and use an input value which can either be the input matrix, or the output of another matrix. This would also accomodate that certain functional methods like decoupler can be executed both, on the gene expression matrix, and the differential expression results.
> 
> ```shell
> analysis_name,method,source,limma_use_voom,signature_file
> # run limma on gene expression matrix
> limma_rnaseq,limma,GEX_MATRIX,true,
> # run decoupler on limma result. `source` references another "analysis name"
> decoupler_rnaseq_de_progeny,decoupler,limma_rnaseq,,progeny.tsv
> decoupler_rnaseq_de_collectri,decoupler,limma_rnaseq,,collectri.tsv
> # run decoupler on gene expression matrix
> decoupler_rnaseq_progeny,decoupler,GEX_MATRIX,,progeny.tsv
> # run limma on decoupler results (if decoupler is run on GEX matrix, it generates separate values for each sample, we still need to do stats)
> limma_progeny,limma,decoupler_rnaseq_progeny,false,
> ```

Actually, even the currently already implemented code has two main chain routes, which assembles to the cases you mention:
1. differential -> de results -> functional
2. (normalization or not) -> matrix -> functional
The first is the case for deseq2+gprofiler2 for example, and the second for deseq2+gsea

What about making this two chain route options explicit?
Something like:
```
analysis_name,study_type,matrix_method,differential_method,functional_method,functional_input,limma_use_voom
deseq2_gprofiler2,rnaseq,,deseq2,gprofiler2,differential,  # get de results from deseq2_diff and run gprofiler2
deseq2_gsea,rnaseq,deseq2,,gsea,matrix,                    # get normalized matrix from deseq2_norm and run gsea
limma_gsea_affy,affy,norm,,gsea,matrix,false               # get default normalized matrix from validator, and run gsea
limma_decoupler,rnaseq,,limma,decoupler,differential,true  # run decoupler on limma-voom results
decoupler,rnaseq,raw,,decoupler,matrix,                    # run decoupler on input gene expression data
```

suzannejin:
> Hi @pinin4fjords and @suzannejin,
> 
> I like the direction this is going!
> 
> > I propose we simplify this by making the toolsheet directly represent pipeline parameters.
> > Each row represents a complete set of pipeline parameters
> 
> This sounds a whole lot like nextflow profiles. With the difference being that multiple "profiles" can be run in a single pipeline invokation?
> 

Exactly! Although, another thing that we would like to explore at some point is the possibility of integrating methods of the same class (for example different DE results, along the line of this: https://pmc.ncbi.nlm.nih.gov/articles/PMC11614847/). Then the value of multiply running methods would go beyond simply parallelize at run time, or beyond benchmark individual methods.

pinin4fjords:
[7996f76](https://github.com/nf-core/differentialabundance/pull/448/commits/7996f764468f76b45a723f3738cd54684b64871d) shows what I mean with the grouping and transposing.

Some improvement possible, but shows how I think this should go. A function can be used to simplify metadata dynamically before certain operations, alllowing the use of groupTuple to collapse channels and prevent duplicated analysis. The deduplicated metas store the 'analysis_name' values, so a transpose operation can be used to restore channel structure after the fact. 

This is still a lot clearer I think

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -304,6 +304,57 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
+    // First collect all differential parameters
+    params.each { key, value ->
+        if (key.startsWith('differential_') || key.startsWith('limma_') || key.startsWith('deseq2_')) {
+            differentialParams[key] = value
+        }
+    }
+
+    // Then collect functional parameters based on method
+    if (params.functional_method == 'gprofiler2') {
+        // For gprofiler2, include both base and differential params
```
suzannejin, line 355: i guess this is because of the fact that gprofiler2 module needs the functional params but also some differential params, as reflected in modules.config. May you add comment too?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -304,6 +304,57 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
+    }
+
+    // Base params are everything not in other sets
+    params.each { key, value ->
+        if (!differentialParams.containsKey(key) && !functionalParams.containsKey(key)) {
+            baseParams[key] = value
+        }
+    }
+
+    // Add base params to both differential and functional
```
suzannejin, line 379: base params is returned, why add it to differential params and functional params too? May you explain or add comment?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -389,6 +440,58 @@ def createToolsheetSchema() {
     return temp_schema.absolutePath
 }
 
+// Simplify meta map to only include parameters relevant for a specific analysis type
+def subsetMeta(meta, analysis_type) {
+    def relevantParams = [:]
+
+    // Define static lists of parameter prefixes for different analysis types
+    def differentialPrefixes = ['differential_', 'limma_', 'deseq2_']
```
suzannejin, line 480: i know that this is only a proof of concept and you are cleaning the code later, I just add a note here for later :)
I think the method specific prefixes could be added dynamically based on the meta.differential_method or meta.functional method. Otherwise relevantParams include params from non relevant methods.

workflows/differentialabundance.nf
```diff
@@ -390,6 +389,27 @@ workflow DIFFERENTIALABUNDANCE {
...
+    // This prevents redundant runs when multiple tool sheet rows require the same analysis
+
+    ch_deduplicated_differential_input = ch_pre_differential_input
+        .map{meta, matrix, diff_method, diff_fc_threshold, diff_qval_threshold, samplesheet, transcript_lengths, control_features, contrasts ->
+            def simplified_meta = subsetMeta(meta, 'differential')
+            [simplified_meta, meta.analysis_name, matrix, diff_method, diff_fc_threshold, diff_qval_threshold, samplesheet, transcript_lengths, control_features, contrasts]
+        }
+        .groupTuple()
+        .map{diff_meta, analysis_names, matrix, diff_method, diff_fc_threshold, diff_qval_threshold, samplesheet, transcript_lengths, control_features, contrasts ->
+            diff_meta.analysis_names = analysis_names
```
suzannejin, line 394: ahhh nice, i like the trick of adding all the analysis names shared by the same diff set, to facilitate later meta expansion while avoiding repeated computation :)

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: Yep, this is because changes in differential params effect gprofiler2, in a way that is not true for GSEA (as currently implemented). Feel free to suggest a rewording?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: So, the way I'm thinking of this, we can iterate on ANY parameter, but we need to control when we can share results across rows, and when we can't. 

By adding base parameters to e.g. the differential set, I ensure that any variation in e.g. the input matrix would prevent sharing of results, even if the differential parameterisation is identical. 

I'll try and add a comment to explain

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -240,167 +304,238 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
+
+    // Base params are everything not in other sets
+    params.each { key, value ->
+        if (!differentialParams.containsKey(key) && !functionalParams.containsKey(key)) {
+            baseParams[key] = value
+        }
     }
 
-    return parsed_params
+    // Add base params to both differential and functional
```
pinin4fjords, line 379:
```suggestion
    // Add base params to both differential and functional. We'll be comparing metas
    // to determine when we can share results across parameter iterations, so we need 
    // to make sure that anything which could lead to differences in e,g. differential 
    // results is recorded. This is the things upstream of differential, and well as 
    //differential parameters themselves.
```

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: True, though there's no cost to that.

workflows/differentialabundance.nf
```diff
@@ -187,245 +152,299 @@ workflow DIFFERENTIALABUNDANCE {
...
-    // collect differential results
+    // Collect differential results
+    // reapplyParams fetches back the study_meta from the paramset
+
+    ch_differential_results = reapplyParams(ABUNDANCE_DIFFERENTIAL_FILTER.out.results_genewise, ch_paramsets_by_name)
+    ch_differential_results_filtered = reapplyParams(ABUNDANCE_DIFFERENTIAL_FILTER.out.results_genewise_filtered, ch_paramsets_by_name)
+
+    // Processed matrices, one pers study/ toolsheet row, so we can reset the meta to the study_meta
+
+    ch_differential_norm = reapplyParams(ABUNDANCE_DIFFERENTIAL_FILTER.out.normalised_matrix, ch_paramsets_by_name)
```
pinin4fjords: I don't like this, by the way, I'm trying to think of something nicer

grst:
> Actually, even the currently already implemented code has two main chain routes, which assembles to the cases you mention:
> 
> 1. differential -> de results -> functional
> 2. (normalization or not) -> matrix -> functional
>    The first is the case for deseq2+gprofiler2 for example, and the second for deseq2+gsea
> 
> What about making this two chain route options explicit? Something like:
> 
> ```
> analysis_name,study_type,matrix_method,differential_method,functional_method,functional_input,limma_use_voom
> deseq2_gprofiler2,rnaseq,,deseq2,gprofiler2,differential,  # get de results from deseq2_diff and run gprofiler2
> deseq2_gsea,rnaseq,deseq2,,gsea,matrix,                    # get normalized matrix from deseq2_norm and run gsea
> limma_gsea_affy,affy,norm,,gsea,matrix,false               # get default normalized matrix from validator, and run gsea
> limma_decoupler,rnaseq,,limma,decoupler,differential,true  # run decoupler on limma-voom results
> decoupler,rnaseq,raw,,decoupler,matrix,                    # run decoupler on input gene expression data
> ```

That could work, although I could also see it being useful to run the differential expression on the result of the functional analyses. In that case, we'd need a `differential_input` column too. 


Note that given the number of possible arguments, redundancies in the toolsheet could be a lot uglier than the example above. There's ~50 possible parameters that can be passed to differential methods at the moment.

pinin4fjords:
> > Actually, even the currently already implemented code has two main chain routes, which assembles to the cases you mention:
> > 
> > 1. differential -> de results -> functional
> > 2. (normalization or not) -> matrix -> functional
> >    The first is the case for deseq2+gprofiler2 for example, and the second for deseq2+gsea
> > 
> > What about making this two chain route options explicit? Something like:
> > ```
> > analysis_name,study_type,matrix_method,differential_method,functional_method,functional_input,limma_use_voom
> > deseq2_gprofiler2,rnaseq,,deseq2,gprofiler2,differential,  # get de results from deseq2_diff and run gprofiler2
> > deseq2_gsea,rnaseq,deseq2,,gsea,matrix,                    # get normalized matrix from deseq2_norm and run gsea
> > limma_gsea_affy,affy,norm,,gsea,matrix,false               # get default normalized matrix from validator, and run gsea
> > limma_decoupler,rnaseq,,limma,decoupler,differential,true  # run decoupler on limma-voom results
> > decoupler,rnaseq,raw,,decoupler,matrix,                    # run decoupler on input gene expression data
> > ```
> 
> That could work, although I could also see it being useful to run the differential expression on the result of the functional analyses. In that case, we'd need a `differential_input` column too.
> 
> Note that given the number of possible arguments, redundancies in the toolsheet could be a lot uglier than the example above. There's ~50 possible parameters that can be passed to differential methods at the moment.

Yeah, I REALLY think a direct mapping from toolsheet columns to parameters is the simplest path forwards right now, even if we can refine the backend parts a bit.

grst:
> Yeah, I REALLY think a direct mapping from toolsheet columns to parameters is the simplest path forwards right now, even if we can refine the backend parts a bit.

I don't disagree with this. Both the suggestion to add a `input` column or a `functional_input` column are fully compatible with being a parameter.

pinin4fjords:
> > Yeah, I REALLY think a direct mapping from toolsheet columns to parameters is the simplest path forwards right now, even if we can refine the backend parts a bit.
> 
> I don't disagree with this. Both the suggestion to add a `input` column or a `functional_input` column are fully compatible with being a parameter.

Ahh yes, see what you mean now

pinin4fjords:
@suzannejin I'm going to need to put most of my time on other projects soon, which is going to slow me down on this. Would you be up for developing on from here? 

You can merge this into your branch and go from there, or just apply the same philosophy if you prefer. This branch does kind of work (though you have to turn off `immundeconv` until we fix that), it just needs the last bits re-added to do with the reporting etc.

Key things from this POC I think we need to keep:

1. Everything from channels via `filter` etc (no `if (params.x`). 
2. modules.config just pulls from meta.param rather than params.param
3. toolsheets represent single iterations of pipeline configuration - no 'special' columns

I can come back and refine a bit, but I'll be slower, so depends how hurried you are!

suzannejin:
> @suzannejin I'm going to need to put most of my time on other projects soon, which is going to slow me down on this. Would you be up for developing on from here?
> 
> You can merge this into your branch and go from there, or just apply the same philosophy if you prefer. This branch does kind of work (though you have to turn off `immundeconv` until we fix that), it just needs the last bits re-added to do with the reporting etc.
> 
> Key things from this POC I think we need to keep:
> 
> 1. Everything from channels via `filter` etc (no `if (params.x`).
> 2. modules.config just pulls from meta.param rather than params.param
> 3. toolsheets represent single iterations of pipeline configuration - no 'special' columns
> 
> I can come back and refine a bit, but I'll be slower, so depends how hurried you are!

Sure! Thank you for the work! I can take the relay baton ;) 
Actually like this dynamic :)

suzannejin (approved):

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -240,167 +336,238 @@ def methodsDescriptionText(mqc_methods_yaml) {
...
-    def toolsheet_path = getToolsheetPath()
+// Create a temporary schema file for toolsheet validation
+// This schema is derived from the pipeline's nextflow_schema.json by:
+// 1. Reading the toolsheet headers to determine which fields to include
+// 2. Extracting only those properties from the pipeline schema
+// 3. Creating a schema that allows for an array of objects
+// 4. Making analysis_name the only required field
+// 5. Adding meta fields to all properties
+// @return The absolute path to the temporary schema file
+def createToolsheetSchema() {
```
suzannejin, line 406: How easy is to maintain this structure of writing temporary schemas given that they can change with nf-schema updates?
Since we are gathering all the params into paramset anyway, I would rather do the validation at that point. 
What do you think @pinin4fjords @mirpedrol

suzannejin: Then, we should also run validateParameters() on the paramset instead of through UTILS_NFSCHEMA_PLUGIN subworkflow

pinin4fjords:
I saw you approved @suzannejin - do you want this merged?

suzannejin:
> I saw you approved @suzannejin - do you want this merged?

Actually let me first work on another branch starting from this one, and I will see :)

Issue #386, opened 2024-12-10
How to define complex contrasts?


grst:
### Description of feature

Follow up of #377. The goal is to find a consensus on how we want users to specify more complex contrasts based on linear model coefficients. 

## Background

Typically linear (and other) models rely on a design matrix for fitting the model and on contrast vectors for performing comparisons. 
A convenient way of constructing such a design matrix is to use a Wilkinson formula. For instance, the formula `~ treatment + response` with the following data frame...

| patient_id | treatment | response |
|--|--|--|
| 1 | drugA | progression |
| 2 | drugA | response | 
| 3 | drugA | stable_disease
| 4 | drugB | response | 
| 5 | drugB | stable_disease |

... would result in the following design matrix

| Intercept | treatmentdrugB | responseresponse | responsestable_disease |
|--|--|--|--|
| 1 | 0 | 0 | 0 | 
| 1 | 0 | 1 | 0 | 
| 1 | 0 | 0 | 1 | 
| 1 | 1 | 1 | 0 | 
| 1 | 1 | 0 | 1 | 

A comparison can be specified using a contrast vector that has the same columns as the design matrix. For instance, to compare `response` against `progression`, we'd need the following contrast vector

| Intercept | treatmentdrugB | responseresponse | responsstable_disease |
|--|--|--|--|
| 0 | 0 | 1 | 0 |

For comparing `response` against `stable_disease`
| Intercept | treatmentdrugB | responseresponse | responsstable_disease |
|--|--|--|--|
| 0 | 0 | 1 | -1 |

A handy R library for exploring design matrices is the R package/Shiny app  [ExploreModelMatrix](https://bioconductor.org/packages/release/bioc/vignettes/ExploreModelMatrix/inst/doc/ExploreModelMatrix.html). 

Below I showcase two options to define complex contrasts, namely limma's `makeContrasts` and glmGamPoi's `cond`.
While both originate from a specific method, they are method agnostic as they generate a numeric contrasts vector. 

## Limma's makeContrasts function

```r
design = model.matrix(~ treatment + response, data=df) # where df is the table from above
makeContrasts(
    response_response_vs_progression = responseresponse,
    response_response_vs_stable = responseresponse - responsestable_disease,
    levels=design
) 
#                         Contrasts
# Levels                   response_response_vs_progression response_response_vs_stable
#  Intercept                                             0                           0
#  treatmentdrugB                                        0                           0
#  responseresponse                                      1                           1
[...]

Issue #363, opened 2024-12-02
Support for linear mixed effects models via DREAM


grst:
### Description of feature

In #325/#339, an attempt was made to add support for limma mixed models. However, I believe this approach is not ideal for the following reasons
 * Limma supports only one random effect variable
 * It requires to manually combine multiple experimental factors into a single variable, therefore making it cumbersome to use

Alternatively, I propose to add support for a linear mixed effects model that generalizes to an arbitrary number of random effects and allows to specify random effect variables using Wilkinson formulas, e.g. `~ timepoint + (1 | patient_id)`. 

An obvious candidate would be [DREAM](https://gabrielhoffman.github.io/variancePartition/articles/dream.html#dream-analysis) which is a generalization of limma/voom for random effects. It's certainly a good choice for gene expression data as it also does the empirical bayes shrinkage of gene dispersions. 

However, there's also other models that could be condidered, e.g. [MRMM](https://github.com/openpharma/mmrm), or plain [lme4](https://cran.r-project.org/web/packages/lme4/vignettes/lmer.pdf) models. 

Depends on #362 for the Wilkinson formulae
Mixed effects models are also a way of addressing technical replicates -> fixes https://github.com/nf-core/differentialabundance/issues/220

CC @apeltzer @tschwarzl @nschcolnicov @atrigila @alanmmobbs93

grst:
IMO DREAM is the prime candidate for mixed effect models on expression data. 

- [ ] create module for [DREAM](https://gabrielhoffman.github.io/variancePartition/articles/dream.html#dream-analysis). It should work farily similar as the limma module
- [ ] add DREAM to differential subworkflow.

alanmmobbs93:
@nschcolnicov @atrigila @grst @apeltzer The basic module and a preliminary position within the workflow is completed. I'll work on  defining some parameters that can be used to customize the modeling in Dream (like limma and deseq2). Later, we need to update the yml structure to decide how/when to call this module with the formula/contrasts definition. I think it can be used in parallel with deseq2 when there's no random variables associated as well. Just to open debate

grst:
Let's keep this open until it's actually in the pipeline

grst:
This is now merged in `dev` #446

PR #450, opened 2025-04-09
feat: allow usage of strings for `makeContrasts` in `DREAM`

atrigila:
Closes: https://github.com/nf-core/differentialabundance/issues/386
Closes: https://github.com/nf-core/differentialabundance/issues/451
Closes: #363

## PR checklist


docs/usage.md
```diff
@@ -197,6 +197,8 @@ The necessary fields in order are:
 
 You can optionally supply:
 
+- `formula` - A string representation of the model formula. It is used to build the design matrix.
+- `string` - An optional explicit literal contrast string (e.g., "treatmenthND6 - treatmentmCherry") that is passed directly to `limma::makeContrasts()` in `VARIANCEPARTITION_DREAM`. This field provides full control for complex designs. Requires `formula`.
```
pinin4fjords: Could you expand this to illustrate what you might want to do here above and beyond the basic case?

pinin4fjords (commented):
I think I see where this is going, makes sense. Could you add a test to illustrate how this would work, something beyond what would be achieved with the current reference/ target setup?

docs/usage.md
grst: Agreed. A good example could be interaction terms, e.g `~ timepoint * treatment`. 

It could also be worth to mention the [`ExplorModelMatrix` shiny app](https://bioconductor.org/packages/release/bioc/vignettes/ExploreModelMatrix/inst/doc/ExploreModelMatrix.html) that helps understanding complex designs and building such contrast vectors.

atrigila: Thank you for your feedback. I added a better explanation for the use case. I also made a PR to shinyngs to change what's required from `validatefomcomponents`. I now also included an interaction case in the tests of `VARIANCEPARTITION_DREAM` https://github.com/nf-core/modules/pull/8231. Once those are merged I can update this PR.

atrigila: Once we get the update of [shinyngs ](https://github.com/bioconda/bioconda-recipes/pull/55567) the `VALIDATOR` module will be able to process the files with `make_contrasts_str` and we will make the dream test with different types of contrast.

atrigila:
A test case with DREAM, `formula` and `make_contrasts_str` was added and it works well. However, I noticed a few issues when I tried to extend the code to more complex scenarios.

So far the `VALIDATOR` R code attempts to [automatically infer blocking variables from the formula](https://github.com/pinin4fjords/shinyngs/blob/d9152985cbc4f537fed107cb9c25be1f4590ac9f/R/accessory.R#L961-L964) by extracting all terms and removing the main comparison variable. This approach worked for simple additive formulas like `genotype + treatment`, where it was clear which terms were independent covariates and also when the 'comparison' field was present. 

However, it fails when the formula contains interaction terms, such as `genotype * treatment`. In these cases the formula expansion creates composite interaction terms that make it difficult to cleanly separate blocking variables from main effects based only on the formula structure. As a result, the automatic detection could mistakenly treat interaction terms as blocking variables, leading to incorrect design matrices. In addition, the[ 'comparison' field is not necessary, as users can supply 'make_contrasts_str"](https://github.com/nf-core/differentialabundance/issues/451). Given this is `NULL` then blocking variables are considered to be all the terms of the formula.

To solve this, I will further modify the `VALIDATOR` R code so that users must now explicitly specify blocking factors through the `blocking_factors` field in the YAML file. This approach avoids ambiguity.

grst:
Just left a comment in https://github.com/nf-core/differentialabundance/issues/451#issuecomment-2834065665 and I think it's related. It could make things easier if formula and blocking_factor were mutually exclusive.

pinin4fjords:
> Just left a comment in [#451 (comment)](https://github.com/nf-core/differentialabundance/issues/451#issuecomment-2834065665) and I think it's related. It could make things easier if formula and blocking_factor were mutually exclusive.

Yeah, I was thinking along these lines as well. It's too complicated when we have overlapping requirements for the formula and non-formula modes.

So @grst , we'd end up with, either:

- You supply everything manually- formula, contrast_string
- OR you stay old-school and supply variable, ref, target, blocking

Yes?

grst:
> So @grst , we'd end up with, either:
> 
> * You supply everything manually- formula, contrast_string
> * OR you stay old-school and supply variable, ref, target, blocking

exactly

atrigila:
Tests are failing because we added the restricted option that `Contrast id 'phenotype_uninvolved_lesional' with 'comparison' must not have 'formula' or 'make_contrasts_str'`.

All of the `contrast_yml` files that are in test-datasets (and that are used in the nf-tests and in the tests) include `'comparison'` and `'formula'`. Example: https://github.com/nf-core/test-datasets/blob/differentialabundance/testdata/formula_contrasts/GSE50790_contrasts.yaml

The next steps should involve:
- Either replacing the `contrast_yml` files for version that comply with out new validation terms, such as the 'old' yaml files: https://github.com/nf-core/test-datasets/blob/differentialabundance/testdata/GSE50790_contrasts.yaml
- Modifying the yaml files that include `formula` so that they also include ` 'make_contrasts_str'` and not comparison.

docs/usage.md
```diff
@@ -175,30 +175,44 @@ contrasts:
...
+
+- `id` - an arbitrary identifier, will be used to name contrast-wise output files
+- `comparison`(respectively):
+- - `variable` - which column from the observations information will be used to define groups
+- - `reference` - the base/ reference level for the comparison. If features have higher values in this group than target they will generate negative fold changes
+- - `target` - the target/ non-reference level for the comparison. If features have higher values in this group than the reference they will generate positive fold changes
+- `blocking_factors` - Any additional variables (also observation columns) that should be modelled alongside the contrast variable
+- `exclude_samples_col` and `exclude_samples_values` - the former being a valid column in the samples sheet, the latter a list of values in that column which should be used to select samples prior to differential modelling. This is helpful where certain samples need to be excluded prior to analysis of a given contrast.
+
+Alternatively, the YAML contrasts also supports formula based model definitions for tools such as `VARIANCEPARTITION_DREAM`:
```
suzannejin, line 188: do this formula based model work on other tools like limma and deseq2 now?
If not please add warning

docs/usage.md
```diff
@@ -175,30 +175,44 @@ contrasts:
...
 
 ```yaml
 contrasts:
   - id: condition_control_treated
     formula: "~ condition"
-    comparison: ["condition", "control", "treated"]
+    make_contrasts_str: "conditiontreated"
```
suzannejin, line 194: It is not very clear to me what is the exact format for `make_contrasts_str`, maybe expand a bit more the explanation and its connection with `formula`?

docs/usage.md
atrigila: Thanks for reviewing! It only works for DREAM now but I plan to follow with limma. I clarified this in the docs now.

docs/usage.md
atrigila: Thanks for your feedback. I added more clarification on this. It's true that you have to know beforehand the model matrix to write the `makeContrast` string. There are a few exploratory tools for that which we added.

workflows/differentialabundance.nf
```diff
@@ -187,8 +187,13 @@ workflow DIFFERENTIALABUNDANCE {
             .map { entry ->
                 def yaml_file = entry[1]
                 def yaml_data = new groovy.yaml.YamlSlurper().parse(yaml_file)
-                yaml_data.contrasts.collect { contrast ->
-                    tuple('id': contrast.comparison[0])
+                yaml_data.contrasts.collect { constrast ->
```
pinin4fjords: Typo: 'constrast'

workflows/differentialabundance.nf
```diff
@@ -187,8 +187,13 @@ workflow DIFFERENTIALABUNDANCE {
...
+                    if (constrast.containsKey('formula')) {
+                        tuple('id': constrast.id)
```
pinin4fjords: Would it not be simpler to mandate an ID in either case?

docs/usage.md
```diff
@@ -175,30 +175,58 @@ contrasts:
...
-- - `reference` - the base/ reference level for the comparison. If features have higher values in this group than target they will generate negative fold changes
-- - `target` - the target/ non-reference level for the comparison. If features have higher values in this group than the reference they will generate positive fold changes
+- `formula` - A string representation of the model formula. It is used to build the design matrix.
+- `make_contrasts_str` - An explicit literal contrast string (e.g., "treatmenthND6 - treatmentmCherry") that is passed directly to [`limma::makeContrasts()`](https://rdrr.io/bioc/limma/man/makeContrasts.html) in `VARIANCEPARTITION_DREAM`. The parameter names must be syntactically valid variable names in R (see [`make.names`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/make.names.html)). This field provides full control for complex designs. Requires `formula`.
 
-You can optionally supply:
+> [!WARNING]
+> Formula-based contrasts are currently only supported by `VARIANCEPARTITION_DREAM`. They **do not work** with tools like `DESEQ2` or base `LIMMA` implementations yet.
```
pinin4fjords, line 206: I guess this is soon to change?

docs/usage.md
```diff
@@ -175,30 +175,58 @@ contrasts:
...
-    comparison: ["condition", "control", "treated"]
+    make_contrasts_str: "conditiontreated"
   - id: condition_control_treated_blockrep
     formula: "~ condition + replicate"
-    comparison: ["condition", "control", "treated"]
+    make_contrasts_str: "conditiontreated"
 ```
 
-The necessary fields in order are:
+The necessary fields in order are::
```
pinin4fjords: Double colon

pinin4fjords (commented):
Looking good, some minor points

docs/usage.md
atrigila: Yes, but for the moment we'll keep this way because I am not sure if the new implementations will make it to next release.

workflows/differentialabundance.nf
atrigila: I initially changed the logic to always use `contrast.id`, which would be what makes sense for formulas. However, that broke the maxquant test, because it relies only on `id = comparison[0]`.  To fix this, I had to reintroduce the conditional fallback. This ensures backward compatibility for maxquant and other cases where id isn't explicitly set. Ideally, we’d enforce id for all contrast entries, but for now, this preserves support for existing usage patterns.

pinin4fjords: Maybe add a comment to this effect, and link to the relevant line for maxQuant? Then we might remember to go and fix up maxquant in future so this can be simplified.

pinin4fjords (approved):
OK, _think_ this is good.

Issue #458, opened 2025-05-09
Support complex contrasts and models in limma


grst:
### Description of feature

Complex contrast support (specification of formula + model string) has been added to DREAM in #450. 

Let's do the same with limma.

atrigila:
Closed via #462

Issue #275, opened 2024-05-24
Provide a parameter to disable rmarkdown report creation in the differential abundance pipeline


shreya-sharma20:
### Description of feature

When analyzing a large number of samples using the differential abundance pipeline, the markdown notebook compilation step fails due to memory issues.

- Error message: "Error: pandoc document conversion failed with error 9" 
- Process causing this error: NFCORE_DIFFERENTIALABUNDANCE:DIFFERENTIALABUNDANCE:RMARKDOWNNOTEBOOK

For now, I have increased the memory for this process. 

Feature Request: It'll be helpful if there is a parameter to disable this process if reports are not required.

megadesk:
I am having the same problem when analyzing a large number of samples, would please share how you increased the memory for this process to solve the problem. I have tried running the job on an very large EC2 instance but that did not solve the problem. I would like to have the study.html file that is not being created.

dshechter:
Additional suggestion to increase the RAM allocated for this process, e.g. 

```
process {
    withName: 'NFCORE_DIFFERENTIALABUNDANCE:DIFFERENTIALABUNDANCE:RMARKDOWNNOTEBOOK' {
        memory = 200.GB
    }
}
```

Could work with less than 200, just haven't tried it.

rcorfixs:
I have been running jobs using the attached workflow, and one of my workflows has been struggling to finish as it continues to encounter this error due to there being many samples to analyze. This study with 45 samples hasn't successfully finished the differential abundance pipeline, depite having 256 GBs of memory and 64 CPUs per task when running on an HPC. I've run jobs that have anywhere from 5 to 35 samples in the study, and those compile just fine while only having 24 GBs of memory and 8 CPUs per task. If anyone has any suggestions or potential fixes, I'd love to hear your response.
![RNAseq_workflow](https://github.com/user-attachments/assets/def95301-e429-426f-8be2-f07952b5515f)

atrigila:
Closed via #462

Issue #424, opened 2025-01-22
Handle Rmarkdown reports for new methods


suzannejin:

atrigila:
Hi @suzannejin! I wanted to check in regarding this issue, have the new methods already been handled, or could you share which ones are still pending?

Thanks!

suzannejin:
> Hi @suzannejin! I wanted to check in regarding this issue, have the new methods already been handled, or could you share which ones are still pending?
> 
> Thanks!
> 

Hey! @caraiz2001 is planning to make the propr method (which is in the differential subworkflow) work at pipeline level.

Issue #369, opened 2024-12-03
Consider splitting quarto reports into QC and results


grst:
### Description of feature

I feel the current quarto report addresses two different audiences:
 1) the bioinformatician who can assess the more technical information and QC metrics (e.g. was sample normalization appropriate, PCA, ...)
 2) the molecular biologists interpreting the results (list of DE genes, pathway scores etc)

IMO it would make sense splitting the report avoid overloading users with information to reduce the size of individual reports #106

delfiterradas:
Working on branch: https://github.com/delfiterradas/differentialabundance/tree/quarto

grst:
I think this can go into `dev` directly, and should be kept minimal (i.e. no other changes to the report than splitting). 

Once we have the split reports in the `dev` branch, we can successively move over features we have in the `dev_reports` branch.

atrigila:
This ticket can be closed as it was decided on #535 that we would not proceed with this at the moment.

PR #462, opened 2025-05-23
Enable `LIMMA` handling formulas

atrigila:
Closes https://github.com/nf-core/differentialabundance/issues/458
Closes https://github.com/nf-core/differentialabundance/issues/275

This PR is a pipeline proof of concept for enabling `limma` to handle formula-based contrasts. It updates the workflow to accommodate contrast definitions using formula and comparison instead of variable, reference, and target.

Several downstream modules (e.g., `GSEA_GSEA`, `SHINYNGS_APP`) assume that `meta.variable` is always present. Since this is no longer guaranteed when using formula-based contrasts, this PR adds logic to selectively filter or route channels only when variable is available.

I also introduce the possibility to skip the final reports. The original R Markdown assumes that every “grouping” variable you ask it to plot (whether coming from the contrasts file’s `variable` column or from PCA‐derived informative_variables) actually exists as a named column in your sample metadata. When you supply a contrast whose `variable` entries are blank (as is the case of formula-based contrasts, or when PCA yields no valid metadata factors), the code still tries to subset `observations[, informative_variables]`, which ends up including empty strings or names that aren’t real columns. R then throws “undefined columns selected” because you can’t pull any non‐existent column out of your data frame. In short, the notebook expects non‐empty, valid metadata column names at every step but fails when those lists contain blanks or missing names. Therefore, the modification of the R reports is left for other more specific tickets such as: 

https://github.com/nf-core/differentialabundance/issues/424
https://github.com/nf-core/differentialabundance/issues/369

## PR checklist

- [x] Usage Documentation in `docs/usage.md` is updated.


atrigila:
After this draft POC is reviewed, then I can update the upstream subworkflows / modules in nf-core.

docs/usage.md
```diff
@@ -440,6 +439,8 @@ process {
 
 You will not get the final reporting outcomes of the workflow, but you will get the differential tables produced by DESeq2 or Limma, and the results of any gene seta analysis you have enabled.
 
+We have also added a dedicated pipeline parameter, `--skip_reports` that allows you to skip only the RMarkdown notebook and bundled report while leaving other reporting processes active. The `RMARKDOWNNOTEBOOK` process assumes that every grouping variable you pass to it (from the contrasts file’s variable column or PCA-derived informative_variables) exists as a valid, named column in your sample metadata. If you know your metadata or contrasts might be incomplete or non-standard (such as using formula-based yaml files), the you can use this flag to skip these steps.
```
pinin4fjords, line 469: This seems a bit awkward.

I wonder if we should just have a global parameter for supplying grouping variables for reporting purposes, for when they can't be derived from contrasts?

Then, workflow logic is set up to only do reporting in the case of non-formula-based contrasts, or when global grouping variable are supplied. If you want a report you do one or the other.

We could still have the report-skipping for convenience (I think it might even have been asked for), but it doesn't need to be tied up with this stuff.

atrigila: I agree, and I did look into modifying the report, but it turned out to be more complex than initially expected. I believe this warrants a separate ticket, as it may require rethinking and discussing the overall report structure with the rest of the dev team.

atrigila:
There have been some changes in the subworkflow structure for `abundance_differential_filter` and `differential_functional_enrichment` in the toolsheet PR, so I will wait for that to be merged in order to make the changes.

docs/usage.md
grst: > I wonder if we should just have a global parameter for supplying grouping variables for reporting purposes, for when they can't be derived from contrasts?

Makes sense! 

I also see the point that reporting is a whole separate topic to be worked on. What's your preferred way how to proceed here @pinin4fjords? Leave `--skip_reports` for now and open a ticket for reporting, or implement a parameter with grouping variable right now (but keeping it as simple as possible as reporting might change down the road anyway).

atrigila:
This is ready for review. Changes to upstream nf-core subworkflows are required but I'd like to get your agreement if this POC looks fine.

workflows/differentialabundance.nf
```diff
@@ -134,13 +134,14 @@ workflow DIFFERENTIALABUNDANCE {
             def yaml_data = new groovy.yaml.YamlSlurper().parse(yaml_file)
             yaml_data.contrasts.collect { contrast ->
                 if (contrast.containsKey('formula')) {
-                    tuple(meta, contrast.id)
+                return null
```
suzannejin, line 137: could you comment here why remove formula based contrast from the channel?

workflows/differentialabundance.nf
```diff
@@ -683,11 +684,28 @@ workflow DIFFERENTIALABUNDANCE {
...
+
+    // Filter by those channels containing 'variable', regroup
+    ch_contrasts_filtered = ch_contrasts
+        .transpose()
+        .filter { meta, contrast, variable, reference, target, formula, comparison ->
+            variable?.trim()
+        }
+        .groupTuple()
+
+    ch_differential_results_filtered = ch_differential_results
```
suzannejin: there is already a `ch_differential_results_filtered` channel allocating those significant DE entries.

workflows/differentialabundance.nf
atrigila: The channel `ch_contrasts_variables_from_yml` obtains the `variable` from the contrasts file. When contrasts are defined in a yaml file with the following structure (the "formula + make_contrasts_str"):
```
contrasts:
  - id: interaction_genotype_treatment
    formula: "~ treatment * genotype"
    make_contrasts_str: "genotypeWT.treatmentTreated"
```
In this case, while the id exists, it does not tell us what the variable is.

workflows/differentialabundance.nf
atrigila: Thank you for your observation. I have renamed this channel to avoid confusion.

docs/usage.md
grst: @pinin4fjords, we need a decision pls how to move forward with this point. Is it ok to merge this with the skip flag and work on reports later separately?

workflows/differentialabundance.nf
```diff
@@ -683,11 +684,28 @@ workflow DIFFERENTIALABUNDANCE {
...
+        .transpose()
+        .filter { meta, contrast, variable, reference, target, formula, comparison ->
+            variable?.trim()
+        }
+        .groupTuple()
+
+    ch_differential_results_with_variable = ch_differential_results
+        .filter { meta, contrast, results ->
+            contrast.variable?.trim()
+        }
```
pinin4fjords: would a transpose, filter, and re-group after the join below be easier?

workflows/differentialabundance.nf
```diff
@@ -798,21 +816,22 @@ workflow DIFFERENTIALABUNDANCE {
         }
 
     // Render the final report
+    if (!params.skip_reports) {
```
pinin4fjords, line 809: Maybe a silly question, but why does this require an explicit skip? Why not just have channel structure dictate whether this happens or not?

atrigila: The reason is that I was trying to solve this ticket as well: https://github.com/nf-core/differentialabundance/issues/275 which asks for an explicit parameter.

workflows/differentialabundance.nf
atrigila: We need to filter upstream, before the join, because otherwise we would be joining on irrelevant data (e.g., contrasts missing a variable).

workflows/differentialabundance.nf
pinin4fjords: Would things currently break if you didn't select to skip the reporting, and contrasts were incompatible with the reporting? If so, I don't think this option is sufficient, the channel structure should inherently prevent that scenario.

atrigila: I see what you mean, let me try that and I'll get back!

workflows/differentialabundance.nf
pinin4fjords: Yes, but your way you're filtering twice. Wouldn't something like this work?

``` 
    ch_differential_with_contrast = ch_shinyngs
        .join( ch_differential_results.groupTuple() )   // [meta, [meta with contrast], [differential results]]
        .join( ch_contrasts )                           // [meta, [contrast], [variable], [reference], [target], [formula], [comparison]]
        .join( ch_differential_results_with_variable.groupTuple() )   // [meta, [meta with contrast], [differential results]]
        .join( ch_contrasts_filtered )                           // [meta, [contrast], [variable], [reference], [target], [formula], [comparison]]

        .transpose()
        .filter { meta, contrast, variable, reference, target, formula, comparison ->
            variable?.trim()
        }
        .groupTuple()

        .map { meta, meta_with_contrast, results, contrast, variable, reference, target, formula, comparison ->
            // extract the contrast entries from the meta dynamically
            // in this way we don't need to harcode the contrast keys
            def contrast_keys = contrast[0].keySet()
            def contrast_maps = meta_with_contrast.collect { it.subMap(contrast_keys) }
            [meta, results, contrast_maps]   // [meta, [differential results], [contrast maps]]
        }
```

atrigila: I am sorry I attempted this (and some other variations with filtering after the join) and it does not work.

workflows/differentialabundance.nf
atrigila: Thank you for your patience while I was re-checking this. I will try to summarize what I've found and my approach.

The inputs to `RMARKDOWNNOTEBOOK` are:
- `ch_report_input.report_file `-> tuple [meta, template] -> settings + 1 markdown template file
- `ch_report_input.report_params` -> map -> all settings for the report
- `ch_report_input.input_files` -> tuple [map, list of files] -> settings + output files

In the list of files we have in `ch_report_input.input_files` we have the contrast TSV file (obtained from `VALIDATOR`). Previously, in this file it was mandatory that the column `variable` had content, but now since we have formula-based contrasts it might look like something like this:
```
id	variable	reference	target	blocking	formula	make_contrasts_str
treatment_effect_classic_comparison	treatment	Control	Treated			
treatment_plus_genotype					~ treatment + genotype	treatmentTreated
interaction_genotype_treatment					~ genotype * treatment	genotypeWT.treatmentTreated
full_model_with_interactions					~ genotype * treatment * time	genotypeWT.treatmentTreated.time
```
The first example `treatment_effect_classic_comparison` is from the 'old' 'comparison-based' yaml-style and the rest of the rows are formula-based contrasts. 

The contrast TSV file is directly used in several parts of the `Rmd` to define the `informative_variables` that will be used for grouping. 

So I think there is no way that I can pre-filter the channel as the contrast TSV file already contains the 'problematic' rows. What I did was to modify that file in order to 'fill' that variable column with something provided by the users via params (`--report_grouping_variable 'treatment'`). This approach solves the issue. I updated the test case so that now the RMarkdown report is rendered (removed the `--skip_reports` option from the limma test case).

pinin4fjords: Can't we do something like [this](https://github.com/nf-core/differentialabundance/blob/3dd360fed0dca1780db1bdf5dce85e5258fa2253/workflows/differentialabundance.nf#L588), re-creating the contrasts file dynamically, sans the problematic rows?

workflows/differentialabundance.nf
pinin4fjords: Could you describe the errors you encountered doing this?

workflows/differentialabundance.nf
atrigila: In that case we are removing the rows containing the formula-based contrasts and we won't be able to add the external grouping variable via params.

workflows/differentialabundance.nf
atrigila: It's a `ERROR ~ Invalid method invocation call with arguments: [whole new long channel] (java.util.ArrayList) on _closure61 type` for this for example:
```
ch_differential_with_contrast = ch_shinyngs
    .join( ch_differential_results.groupTuple() )
    .join( ch_contrasts )
    .transpose()
    .filter { meta, contrast, variable, reference, target, formula, comparison ->
        variable?.trim() && contrast.variable?.trim()
    }
    .groupTuple()
    .dump(tag:"ch_differential_with_contrast")
```

workflows/differentialabundance.nf
atrigila: I mean, right now we re-creating the contrast file as you suggest but adding a new column which allows for plotting via this new global parameter for supplying grouping variables for reporting purposes, for when they can't be derived from contrasts. What do you think about this?

pinin4fjords: Ahh, I see what you're doing there now.

**If** we keep something like this, I suggest you just make that re-creation of the contrast channel unconditional. Set the variable from the global setting (if defined) within a map, similar to how you're doing, then filter out any rows that still have no variable set after that. That would mean that in the case where a user doesn't skip reporting, and doesn't provide that variable, the reporting process just doesn't happen (because `ch_validated_contrast` will be an empty channel.

**However** this is starting to feel too hacky and complex.

Another thought- should we maybe just allow people to set `variable` even when they're using formulas? Then all we have to do is remove contrast rows that don't have the variable set. If that means we lose all the contrast rows, then no reporting will happen.

workflows/differentialabundance.nf
pinin4fjords: OK, I'm clocking off right now, but I'm going to try to find time to test this out tomorrow.

workflows/differentialabundance.nf
atrigila: I see what you mean. Then I need to remove my current implementation and go back to the previous one (before this commit https://github.com/nf-core/differentialabundance/pull/462/commits/5834f5ee596f6649e22fea914a3fa8cea24f1480) where I populated the `--report_grouping_variable 'treatment'` in the variable column. I only need to add a filter to remove any rows that still have no variable set after that. I'll let you know when I added that.

atrigila: done!

workflows/differentialabundance.nf
pinin4fjords: OK, so both of these filtering steps can be removed, and we can just add the filtering as part of the map below. I checked, and this works:

```
    ch_differential_with_contrast = ch_shinyngs
        .join( ch_differential_results.groupTuple() )   // [meta, [meta with contrast], [differential results]]
        .join( ch_contrasts )                           // [meta, [contrast], [variable], [reference], [target], [formula], [comparison]]
        .map { meta, meta_with_contrast, results, contrast, variable, reference, target, formula, comparison ->
            // extract the contrast entries from the meta dynamically
            // in this way we don't need to harcode the contrast keys
            def paramset_contrast_keys = contrast[0].keySet()
            def indices_of_results_with_variable = meta_with_contrast.findIndexValues { it.variable?.trim() }
            def contrast_maps = indices_of_results_with_variable.collect { i -> meta_with_contrast[i].subMap(paramset_contrast_keys) }
            def filtered_results = indices_of_results_with_variable.collect { i -> results[i] }
            [meta, filtered_results, contrast_maps]
        }
```

pinin4fjords: Since all the contrast channel is doing here is supplying the keys, we don't need to filter it at all. Then we can remove both of these filtering steps, and the segment below becomes just:

```
    ch_differential_with_contrast = ch_shinyngs
        .join( ch_differential_results
            .filter { meta, contrast, results -> contrast.variable?.trim() }
            .groupTuple() 
        )   // [meta, [meta with contrast], [differential results]]
        .join( ch_contrasts )                           // [meta, [contrast], [variable], [reference], [target], [formula], [comparison]]
        .map { meta, meta_with_contrast, results, contrast, variable, reference, target, formula, comparison ->
            // extract the contrast entries from the meta dynamically
            // in this way we don't need to harcode the contrast keys
            def paramset_contrast_keys = contrast[0].keySet()
            def contrast_maps = meta_with_contrast.collect { it.subMap(paramset_contrast_keys) }
            [meta, results, contrast_maps]
        }
```

workflows/differentialabundance.nf
pinin4fjords: I actually PR'd your PR- see https://github.com/atrigila/differentialabundance/pull/2

I believe that much of what you were doing had already been done elsewhere in the workflow, so we can just use those rather than re-filtering. The PR is not quite complete- you'll need to add your injection of the global param val where appropriate, but hopefully you see what I mean.

workflows/differentialabundance.nf
pinin4fjords: see https://github.com/atrigila/differentialabundance/pull/2

atrigila: Thank you, I accepted your PR. It is not clear to me why it breaks the maxquant test, though, as it is simply a refactoring of the previous code that worked.

atrigila: I understand why now, it does make sense, we have a:

```
    ch_shinyngs = ch_paramsets
        .filter{ it.params.shinyngs_build_app }
```
and we in the maxquant config it is defined that:
```
    // Shiny does not work for this datatype
    shinyngs_build_app               = false
```

Then, downstream channels are chained off `ch_shinyngs`:
```
   // Create a channel with the differential results and the corresponding map with
    // the contrast entries
    differential_with_contrast = ch_shinyngs
        .join( ch_differential_results
            .filter { meta, contrast, results -> contrast.variable?.trim() }
            .groupTuple() 
```

workflows/differentialabundance.nf
atrigila: If you agree, I think the injection of the global param will be better to be left for another PR as if we allow people to set variable even when they're using formulas, the changes have to be done when reading the `ch_contrasts`, at this stage:
```
    ch_contrasts = ch_validated_contrast
        .splitCsv ( header:true, sep:'\t' )
        .map{meta, contrast ->
            contrast.blocking = contrast.blocking.replaceAll('^NA$', '')
            if (!contrast.id){
                contrast.id = contrast.values().join('_')
            }
            contrast.formula = contrast.formula?.trim() ? contrast.formula.trim() : null
            contrast.make_contrasts_str = contrast.make_contrasts_str?.trim() ? contrast.make_contrasts_str.trim() : null

            return [meta, contrast, contrast.variable, contrast.reference, contrast.target, contrast.formula, contrast.make_contrasts_str]
        }
        .groupTuple() // [meta, [contrast], [variable], [reference], [target], [formula], [comparison]]
```
        
If that happens, then it has many implications in another downstream processes (such as `GSEA_GSEA`) and not only in the reporting stage. Unless we do something in the `differential_with_contrast` stage, but as you mentioned, it is starting to feel a bit hacky. I am open to suggestions.

pinin4fjords: Yep, I think an MVP to just filter the incompatible rows is fine.

tests/test_maxquant.nf.test.snap
```diff
@@ -234,9 +231,6 @@
                 "plots/qc/genotype_celltype_t1_FoB_PXD043349.limma.mean_difference.png",
                 "plots/qc/genotype_celltype_t1_MZ_fakeBatch_PXD043349.limma.mean_difference.png",
                 "plots/qc/genotype_celltype_t1_t2_PXD043349.limma.mean_difference.png",
-                "report",
-                "report/PXD043349.html",
```
pinin4fjords: Why are these files no longer produced with the Maxquant profile here??

atrigila: I had the same question. See my comment here:
https://github.com/nf-core/differentialabundance/pull/462#discussion_r2185533186

workflows/differentialabundance.nf
atrigila: Then, that's addressed as the current structure achieves that by filtering the rows without variable.

tests/test_maxquant.nf.test.snap
pinin4fjords: Ahh yes. I think we'll have to figure that out before this can be merged

atrigila: done! :)

pinin4fjords: Ahh- nice fix

pinin4fjords (approved):
OK, I think we may choose to polish a bit before release, but looks good.

PR #535, opened 2025-10-14
Split quarto reports into QC and results

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/369

## PR checklist


grst (changes requested):
Feedback:

* the filenames do not match well the contents. I'd suggest `report_exploratory.qmd` and `report_differential.qmd` instead
* h1 titles are "Differential gene abundance report" in both cases. I'd suggest "Exploratory analysis report" and "Differential gene abundance report", respectively. 
* move "contrasts" section from exploratory -> differential report

atrigila (commented):
Can you update the `output.md` (and README and usage if applicable) with the new structure that is generated? Additionally, could you add the HTML files that are now being generated for easier inspection? Thank you.

delfiterradas:
Yes! These are the new HTML files: 

- [SRP254919_report_exploratory.html](https://github.com/user-attachments/files/23009751/SRP254919_report_exploratory.html)

- [SRP254919_report_differential.html](https://github.com/user-attachments/files/23009753/SRP254919_report_differential.html)

grst (approved):
LGTM now

SusiJo:
Hi,

Thanks for the work on this PR! We have a few questions and points we’d like to discuss regarding the report split before it’s merged.

It might be helpful to have a joint discussion on how and where the report should be split — specifically, which sections belong in each part and whether this split should be the default behavior or an optional feature.

As it stands, the current split doesn’t fully align with typical pipeline user needs. Since the two reports don’t reference each other, it can be a bit cumbersome to switch back and forth to get a complete overview of the analysis.

From our perspective, it would make sense for certain sections — for example, the PCA — to also appear in the molecular report, not just the bioinformatic one.

We’d also like to revisit the naming conventions: one report still carries the pipeline name, while the other is called molecular_report. These names don’t clearly reflect the bioinformatic vs. molecular distinction.

Overall, we’d be in favor of making this report split optional rather than enabling it by default.

Thanks again for your efforts — we’re looking forward to discussing this further!

grst:
Hi @SusiJo, 

thanks for your feedback. 

> As it stands, the current split doesn’t fully align with typical pipeline user needs. 

Frankly, I think this is one of the biggest problems with this pipeline and we should discuss this at the hackathon: I have the impression that the views on what this pipeline should be and whom it should serve deviate quite a bit between the different developers. 

> Since the two reports don’t reference each other, it can be a bit cumbersome to switch back and forth to get a complete overview of the analysis

This probably makes sense to address, independent of how reports are split. Creating some sort of index and crossreferencing is certainly a good idea. I would just like to keep the ability to render reports independently in parallel, rather than creating, say, a quarto book. 

> We’d also like to revisit the naming conventions: one report still carries the pipeline name, while the other is called molecular_report. These names don’t clearly reflect the bioinformatic vs. molecular distinction.

The names have been revised. They are now called "report_exploratory" (with everything agnostic of defined contrasts) and "report_differential" (with everyting related to the contrast-based differential analysis). What do you think of this distinction? 

> Overall, we’d be in favor of making this report split optional rather than enabling it by default.

I think one of the main advantages of the new setup is that anyone can pass on custom quarto documents to the pipeline. That makes it easy for everyone to tailor the report to their needs. Given the diverse audience of this pipeline, I'm not even sure we can make one default report that satisfies everyone, but we can still try our best.

SusiJo:
> Hi 
> 
> > Frankly, I think this is one of the biggest problems with this pipeline and we should discuss this at the hackathon: I have the impression that the views on what this pipeline should be and whom it should serve deviate quite a bit between the different developers.
> 
> Yes let's discuss this in more detail at the hackathon! Then we can also talk about cross-referencing the reports and the naming schema.
>
> > The names have been revised. They are now called "report_exploratory" (with everything agnostic of defined contrasts) and "report_differential" (with everyting related to the contrast-based differential analysis). What do you think of this distinction?
> 
> These revised names sound more natural and intuitive, also from a user perspective.

pinin4fjords:
Just to confirm @SusiJo are you happy now?

tests/default.nf.test
```diff
@@ -148,7 +148,7 @@ nextflow_pipeline {
                 outdir        = "$outputDir"
                 contrasts_yml = null
                 contrasts     = "https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/genomics/mus_musculus/rnaseq_expression/SRP254919.contrasts.csv"
-                report_file   = "$projectDir/tests/quarto.qmd,$projectDir/assets/differentialabundance_report.qmd"
+                report_file   = "$projectDir/tests/quarto.qmd,$projectDir/assets/report_differential.qmd"
```
SusiJo: Can this `$projectDir/tests/quarto.qmd` be removed or should it be changed to `$projectDir/assets/report_exploratory.qmd`?

SusiJo:
Hi, 
we definitely require cross-links between the two reports. Currently there is no sample metadata information in the differential_report, like how many samples and genes were used to conduct the differential analysis. It is necessary to jump between the reports to get the full picture of the analysis. Could such links please be added as part of this PR?

Apart from that, QBiC would need the PCA plots in the differential_report (or at least a link). It would be really great if this can be added here.

grst:
I'm afraid the linking part will become a rabbit hole. Let's keep a single report for now and revisit this once we are happy with the overall contents.

Issue #531, opened 2025-10-10
Allow integer sample identifiers


grst:
### Description of the bug

I don't see a good reason not to allow numeric sample identifiers and I don't think it's that uncommon to have some. 

Part of the issue is that csvs are not typed. I think it would be reasonable to accept any type for sample id and just convert it to str internally. 

### Command used and terminal output

```console
--input (../../01_prepare_data/output/1403-0008_rbm_pd/input.tsv): Validation of file failed:
        -> Entry 1: Error for field 'sample' (652111231906): Value is [integer] but should be [string] (Sample name must be provided and cannot contain spaces)
```

atrigila:
Possible options:
1. `lenientMode = true` as in https://github.com/nf-core/demultiplex/blob/1b94ed6c8d5e315e41cf58521a7671ff0da8b9b6/nextflow.config#L431
2. `anyOf` as in https://github.com/nf-core/sarek/blob/20f41d1ce8b7ba296ee22adc71fe2da2ebcae93f/assets/schema_input.json#L37

grst:
I don't think a problem with one column justifies lenientMode for the other ones. `anyOf` sounds good.

PR #581, opened 2025-12-19
Allow sample to be an integer, apply module updates

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/531 and https://github.com/nf-core/differentialabundance/issues/584

Changes:

- Update `variancepartition/limma` to handle integers as sample IDs.
- Update `dream/differential` to handle integers as sample IDs.
- Install latest versions of modules and subworkflows.
- Add tests and testdata to show pipeline function with numeric sample IDs.

## PR checklist


delfiterradas:
These changes would allow the use of numeric sample IDs however, until the changes from https://github.com/nf-core/modules/pull/9562 are reintegrated, `variancepartition/dream` will fail if using a sample ID that is an integer.

grst:
Can we force this to be a character/factor at some point in the pipeline such that this change gets propagated to all models?

I'm afraid that it could be interpreted as a numeric covariate in any of the models which would lead to completely wrong results.

delfiterradas:
> Can we force this to be a character/factor at some point in the pipeline such that this change gets propagated to all models?
> 
> I'm afraid that it could be interpreted as a numeric covariate in any of the models which would lead to completely wrong results.

@grst I have been testing some things out, but I don't think it is possible to force it to be a character within the pipeline since the sample IDs are only validated in the input file but not used within channels by themselves. They are either column or row names in the matrices or samplesheets.

I have tried running the pipeline with numeric sample names with the different modes and `deseq2` seems to be the only one working as is (and getting the same results). I am working on adding a test case to show this.

As for `dream` and `limma`, I think the best option would be to modify the R scripts to make sure they are always transformed to character. As I mentioned in another comment the fix for `dream` is already in progress.

grst:
Thanks for investigating! Can you please create tickets for dream/limma and work on the fixes? 

I still believe this is an unnecessary source of errors and we should consider other formats for passing around the data internally. This (and a possibility to preserve factor levels) is one of the reasons for which I'd prefer SummarizedExperiments, but Jon was against R-specific formats. Another way could be parquet files, which are language agnostic. But that requires a broader discussion and I'll open a separate ticket for that. In the meanwhile, I the pragmatic way forward is to adjust the individual modules.

delfiterradas:
@grst - Both `dream` and `limma` have been updated and now handle integers as sample IDs. I have also added a new test scenario for each of them 🙂

SRP254919.contrasts.csv
atrigila, line 1: Can we move this file into `tests/assets`? Thank you

atrigila: Or I am not sure if this file was used for testing and should be removed?

grst (approved):

pinin4fjords (changes requested):
There are a lot of changes here that don't seem directly related to the issue at hand. Could you confirm that you intended this, and amend the PR description if you did please?

grst:
Essentially this PR is just module updates, which at this point have accumulated other changes including the relevant ones. 

But I'll leave it to @delfiterradas to update the description and possibly changelog to reflect that.

pinin4fjords (approved):

Issue #570, opened 2025-12-04
Replace dendrogram with complex heatmap


delfiterradas:
Migrate the changes from https://github.com/nf-core/differentialabundance/pull/525 to `dev` branch

PR #525, opened 2025-10-07
Add heatmaps with dendrograms

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/506

## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core lint`).
- [ ] Ensure the test suite passes (`nf-test test main.nf.test -profile test,docker`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

Issue #506, opened 2025-09-12
Replace dendrogram by ComplexHeatmap


grst:
### Description of feature

Currently we have plots for a dendrogram colored by all covariates. 

<img width="1032" height="687" alt="Image" src="https://github.com/user-attachments/assets/e536ee4f-d5c8-46c7-9924-ed99197578a5" />

I think it would be more informative, and more compact to reduce this to three heatmaps (VST, Normalized, Raw) that include the same dendrogram clustering, have the colors as column annotations, and show the expression of the $n$ most variable genes that are used to generate the dendrogram. 

Something like this: 

<img width="960" height="960" alt="Image" src="https://github.com/user-attachments/assets/541bf15a-11f9-40a9-9389-1f3aaeeb5553" />

PR #586, opened 2026-01-04
Replace dendogram

antoniasaracco:
This closes https://github.com/nf-core/differentialabundance/issues/570


antoniasaracco:
Hi @grst this migrated the changes from https://github.com/nf-core/differentialabundance/pull/525 to dev branch to include heatmaps with dendrograms. Please let me know your thoughts, thanks!

pinin4fjords:
Could you post before and after HTML files for this please?

antoniasaracco:
Hi @pinin4fjords, 

This is the before where clustering dendrograms were present

image

And here is the after with the heatmaps
image

Thanks!

pinin4fjords:
I see the rationale in how @grst described https://github.com/nf-core/differentialabundance/issues/506 - in that we reduce the number of dendrograms and just show the one with combined annotations. 

But I'm a little unconvinced that the unlabelled heatmap is adding much here, and it's squishing the dendrogram into a very small area at the top of the plot where it's going to be hard to see relationships beyond a trivial number of samples, no?

grst:
you mean feature (i.e. gene) labelling? 

Because I one of the advantages of the new plot is IMO that one can see many variables at the same time in the column annotations. 

image

I'm also not the biggest fan of gene expression heatmaps, but it's a commonly requested plot and maybe useful as a sanity check to spot e.g. normalization issues. Do you think it would be more useful if just some "genes of interest" with corresponding feature labels were shown?

pinin4fjords:
> you mean feature (i.e. gene) labelling?
> 
> Because I one of the advantages of the new plot is IMO that one can see many variables at the same time in the column annotations.
> 
> image
> I'm also not the biggest fan of gene expression heatmaps, but it's a commonly requested plot and maybe useful as a sanity check to spot e.g. normalization issues. Do you think it would be more useful if just some "genes of interest" with corresponding feature labels were shown?

Totally, I love the variables displayed together, it's great. I don't love a massive heatmap of debatable utility compressing the dendrogram so it's hard to see the sample relationships.

grst:
Ok, but what would you suggest going forward?

- remove the heatmap and just keep the dendrogram with all column annotations
- scale up the dendrogram, that heatmap and dendrogram consume ~50% of the vertical space each
- reduce number of features in the heatmap, by focusing on "genes of interest" (which can be the top $n$ genes or genes already specified manually for the volcano plot)

pinin4fjords:
Maybe if you feel this heatmap is useful it could just be in addition to, rather than a replacement of the stand-alone dendrogram? Otherwise yes, scaling up to make sure there's space is good too.

The dendrogram should reflect the features in the heatmap when they are shown together, so restricting to a very small feature set might have undesirable impact on resolution of the dendrogram.

grst:
> The dendrogram should reflect the features in the heatmap when they are shown together, so restricting to a very small feature set might have undesirable impact on resolution of the dendrogram.

fair enough, agree it doesn't make sense to subset

> Maybe if you feel this heatmap is useful it could just be in addition to, rather than a replacement of the stand-alone dendrogram? Otherwise yes, scaling up to make sure there's space is good too.

I would avoid another (partly redundant) plot in an already pretty heavy report. 
So rescaling the heatmap such that the dendrogram gets at least as much space as it had in the previous version without the heatmap would currently be my preferred option.

grst:
@antoniasaracco, Jon has given thumbs up, can you adjust this accordingly please?

antoniasaracco:
Hi @grst @pinin4fjords,

I’ve adjusted the layout to give more vertical space to the dendrogram so it’s easier to read when shown together with the heatmap.

In the image below, the previous version is on the left and the updated version is on the right.

Please let me know if this looks sufficient to you or if you’d prefer the dendrogram to be scaled up further. Thanks!

image

pinin4fjords (approved):
Good enough, thanks!

Issue #484, opened 2025-08-25
Simplify pipeline parameters


grst:
### Description of feature

This pipeline has an enormous number of parameters. Each of them creates a maintenance burden (keeping docs up-to-date, take params into account during modifications) and make it more complicated for users to understand what they need to specify in order to run the pipeline correctly. 

IMO we should try to attempt to reduce the number of parameters once the multi-config toolsheet is fully supported (#470). 

Some candidates for removal:

- [ ]   ~~`study_type` is probably superseeded by the toolsheet?~~
- [x]   `deseq2_cores` -> shouldn't that be inferred from the number of cores assigned to the nextflow task
- [ ]   `--exploratory_palette_name` -> better choose a suitable palette based on the number of categories.
- [ ]   `--grprofiler2_palette_name` -> see above
- [ ]   `--differential_palette_name` -> same
- [x]   `--report_scree` -> just include it. Better to support custom reports instead, see https://github.com/nf-core/differentialabundance/issues/88
- [x]   `--differental_.*_column` -> internally set the appropriate colnames for each method. I don't think there's any point in exposing this to the user. 
- [x]   `--gsea_zip_report` -> an output folder should be enough. Users can zip themselves if they want. 

---
 

- [x]  There are some parameters that have non-intuitive names and could be simplified for users, i.e. --transcript_length_matrix to have a more generic name like --feature_length_matrix

@pinin4fjords LMK what you think

suzannejin:
yes i agree!! many of these parameters, like for example deseq2 specific column values, etc they should not be provided as parameter by the user, but instead fixed inside the pipeline or something.

pinin4fjords:
We should address these one by one. e.g. the fold change logging is method dependent. 

But I agree with the principle, I may have got a little param-happy. Folks can still supply these things via custom config and ext.args.

SusiJo:
Hi, 
I also agree and approve the idea of simplification. There are some parameters that have non-intuitive names and could be simplified for users, i.e. `--transcript_length_matrix` to have a more generic name like `--feature_length_matrix`

PR #615, opened 2026-01-15
Show gene name on volcano plot hover

### LINKED ISSUE #606: Volcano plot should show gene name when hovering over points
### Description of the bug

Currently, the gene id is shown, even when a gene name is available.

Image

### Command used and terminal output

```console

```

### Relevant files

_No response_

### System information

_No response_

atrigila:
Closes #606 
Screenshot 2026-01-15 141323

This requires setting both:
```
differential_feature_name_column: "GENESMBL"
features_name_col: "GENESMBL"

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- [[#xxx](https://github.com/nf-core/differentialabundance/pull/xxx/)] - Volcano plot now shows gene name when hovering over points, review by [@grst](https://github.com/grst).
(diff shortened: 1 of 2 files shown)


grst:
Then I guess my problem was that I didn't set both parameters. 
Are you aware what's their different purposes? They seem redundant and the documentation of both reads very similar. 

Anyway good point to implement the fallback option.

atrigila:
Yes, they are quite similar. My understanding is that one names features at the matrix level, and the other is specific to reporting.  Maybe they can be harmonised into one and could be tracked here as well:https://github.com/nf-core/differentialabundance/issues/484

grst:
It's actually already listed in #484. Jon agreed in principle to address the points there one-by-one. Let's start tackling this. Could you make a follow-up PR to remove/harmonize the `--differential_.*_column` parameters?

grst (approved):

PR #619, opened 2026-01-22
Improve caching behaviour for processes using`VALIDATOR` outputs

### LINKED ISSUE #520: Improve caching behavior
### Description of feature

Right now, when I add or remove a contrast, e.g.

```diff
contrasts:
   - id: BORC_simple_PD_vs_SD
     comparison: ["BORC_simple", "SD", "PD"]
-  - id: BORC_simple_PD_vs_SD_formula
-    formula: ~ 0 + BORC_simple
-    make_contrasts_str: "BORC_simplePD - BORC_simpleSD"
   - id: BORC_simple_PRCR_vs_SD
     comparison: ["BORC_simple", "SD", "CR/PR"]
```

this invalidates the entire cache, since the `VALIDATOR` process reruns on the updated `contrasts.yml` file, and therefore all downstream processes rerun. 

However, basically everything would be available in the cache already, only the final reports would  have to be rerun. 

A simple way around this could be to set the caching behavior of selected processes to `deep` (since we are dealing with rather small input files, it could be worthwhile). Alternatively, maybe the validator process can be refactored to do validation on a per-contrast basis.

atrigila:
Closes #520 

## PR checklist


conf/modules.config
```diff
@@ -162,6 +162,7 @@ process {
     }
 
     withName: CUSTOM_MATRIXFILTER {
+        cache = 'deep'
```
grst: I'd add a comment to each of these entries to explain the *why*.

grst (commented):
This works, and the runtime cost should be minimal since we are not dealing with huge data in differentialabundance. 

Still would be curious what @pinin4fjords  thinks of this approach:
tl;dr: Any change in the inputs causes VALIDATOR to run and invalidates the cache, even if it doesn't affect any downstream processes. Setting `cache = deep` on any process that consumes output of VALIDATOR solves this issue. For more details, see https://github.com/nf-core/differentialabundance/issues/520

pinin4fjords:
This makes me a little nervous, we're assuming a lot about the size of the inputs future folks will apply, and the scale they'll be running at. 

I'd be more comfortable if this was implemented via a profile so folks would do `-profile docker,cache_deep` or similar.

grst:
What I meant was that we are just operating on a fundamentally different scale in this pipeline, then, say, BAM files. 

I can live with the profile, but I think it would be a reasonable default if the caching just worked as expected, especially with large datasets.

pinin4fjords:
> What I meant was that we are just operating on a fundamentally different scale in this pipeline, then, say, BAM files.
> 
> I can live with the profile, but I think it would be a reasonable default if the caching just worked as expected, especially with large datasets.

Yes, but we're still significantly ramping up the overheads (even if from a very low base). How about a compromise, have it on by default, but give me a profile to turn it off and reset `cache` to default? That way the option is handy if the caching causes issues for someone.

grst:
ok, let's do that! Also, if you have any suggestion how to solve this differently, we are happy to give it a try!

pinin4fjords:
> ok, let's do that! Also, if you have any suggestion how to solve this differently, we are happy to give it a try!

I don't have a more cunning plan. The alternative might be to validate different things (matrices, contrasts) separately, but that has its own issues with complexity. This approach is reasonable.

atrigila:
@pinin4fjords @grst added the default profile and an explanation.

pinin4fjords (approved):

Issue #635, opened 2026-02-12
Remove `--gsea_zip_report` parameter


delfiterradas:
> --gsea_zip_report -> an output folder should be enough. Users can zip themselves if they want.

PR #636, opened 2026-02-13
Remove `--gsea_zip_report` parameter

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/635

## PR checklist


grst:
I'm fine removing the zip report entirely, but an other alternative would be to just always include it. I guess the runtime penalty would be minimal. 

@pinin4fjords what do you prefer?

pinin4fjords:
I was really intentional when I added the zip guys, why does it have to be stripped? It's really nice to have something you can distribute with all the data files for the report.

grst:
It doesn't have to, I just don't think it has to be a separate param. Fine with always creating it then. 

I also somewhat think that creating shareable artifacts should be done on a workflow level rather than a tool level, but that would be a bigger discussion.

delfiterradas:
Would it be okay to make the gsea zip by default or should we leave the parameter as is?

pinin4fjords:
I'm not seeing the benefit of stripping the option

grst:
Well, it's one of many mentioned in #484 that could help reduce the number of params. 
I don't see why it needs to be configurable at the param level. Either include it all the time or not at all.

pinin4fjords:
I don't think I'm totally on board with stripping control for no reason. Put options in an 'advanced' section to be ignored if you want, set defaults as you please, but I like the control to be there if I want it.

PR #487, opened 2025-08-29
Add parameter to skip outlier section in report

SusiJo:
## PR checklist

Requested by QBiC

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core lint`).
- [ ] Ensure the test suite passes (`nf-test test main.nf.test -profile test,docker`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

PR #638, opened 2026-02-13
Add option to disable sections in the quarto report

### LINKED ISSUE #637: Include scree plot by default
> `--report_scree` -> just include it.

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/637
Closes #487

- Remove `--report_scree` parameter
- Add `--disable_report_modules` parameter to specify which sections should not be included in the final report (everything will be included by default)

## PR checklist


pinin4fjords:
Can we talk about the param removals guys? We might have chatted previously, but remind me why we care about them being there?

grst:
My motivation is stated in https://github.com/nf-core/differentialabundance/issues/484#issue-3351420947

> This pipeline has an enormous number of parameters. Each of them creates a maintenance burden (keeping docs up-to-date, take params into account during modifications) and make it more complicated for users to understand what they need to specify in order to run the pipeline correctly.

And your response in https://github.com/nf-core/differentialabundance/issues/484#issuecomment-3242204096

> We should address these one by one. e.g. the fold change logging is method dependent.
> 
> But I agree with the principle, I may have got a little param-happy. Folks can still supply these things via custom config and ext.args.

I understand that for some in the list there may be reasons to keep them, but I think it's a valid appraoch to open a PR with the suggestion to remove/modify them and then discuss it. 

---

For this particular one, I don't see why there would be a flag for including the scree plot, but not for other sections. If one wants to customize the quarto report, it's already possible to modify the qmd file and pass it to the pipeline. 

From my perspective that would be enough, but if we want the main report to be more configurable, then I think it would be preferable to have a param like 

```yaml
include_param_modules: ["volcano", "scree", "de_table", "pca", ... ]
```

instead of having one flag for each section.

pinin4fjords:
OK, how about we stick with the principle that you can remove a param if users still have a way of doing the same thing? e.g. if I can use `ext.args` to accomplish the same objective that's fine (so that would allow a lot of the individual tool-wise params to be stripped, which is what I was imagining in that original discussion).

grst:
>  it's already possible to modify the qmd file and pass it to the pipeline.

does that count?

pinin4fjords:
> > it's already possible to modify the qmd file and pass it to the pipeline.
> 
> does that count?

No, if that counted everything would be allowed!

grst:
What bothers me most about this parameter is the inconsistency. It's the only section of the quarto report that you can switch off, while there's no option for the others. 

I don't think that it scales well to add a separate parameter for each section. So I'd propose this instead: 

>  [...] I think it would be preferable to have a param like
>
> ```yaml
> include_param_modules: ["volcano", "scree", "de_table", "pca", ... ]
> ```
> instead of having one flag for each section.

Another advantage is that this would still work with a custom quarto report that might have different "modules".

WDYT?

pinin4fjords:
Yep, I'm up for that, since you're not taking away the control :-)

delfiterradas:
I have added the `--disable_report_modules` parameter where the user can specify a comma separated list of the sections that should not be included in the report. Right now the allowed options are: 

- boxplot
- density
- pca
- scree
- heatmap
- outlier
- upset_plot
- volcano
- de_table
- biotype_plot
- gsea

@grst & @pinin4fjords please let me know what you think and if there are any options that we should add or remove so that I can make the adjustments

atrigila:
I think that this PR will also fulfill this request: https://github.com/nf-core/differentialabundance/pull/487/changes :) just noting this here so that we can take this into account.

assets/differentialabundance_report.qmd
```diff
@@ -1271,7 +1280,7 @@ foo <- lapply(names(p_value_types), function(pvt) {
     )
     cat("\n")
 
-    if (nrow(contrasts) > 1) {
+    if (nrow(contrasts) > 1  && !module_disabled("upset_plot")) {
```
grst: Do you think it would be possible to split this up in chunks such that the `eval=!module_disabled()` flag could be used here, too? 

It would be nice to be more consistent, and probably also the structure of the report would benefit if each "output module" were a different chunk.

delfiterradas: Yes, I will look into this

delfiterradas: @grst I have split this section into two chunks so that the `eval` can be used. This change also means that the upset plot will now be in its own tabset.

However, there are 3 modules (volcano plot, de table and biotype plot) that are still in the same chunk. This is because they are created inside the same loop and render within a single tabset. I don't think it is possible to split them into separate chunks without either duplicating the loop logic or restructuring the code more substantially. Would you be okay with keeping this part as is or should I explore restructuring this section further?

grst (approved):
I can live with the current version.

delfiterradas:
Hi @pinin4fjords, when you have the time would you mind taking a look at these changes? Thank you :)

pinin4fjords (approved):
Thanks @delfiterradas! The new `--disable_report_modules` param retains (and expands) the control we had with `--report_scree`, which was my main concern. Looks good to merge.

PR #644, opened 2026-02-17
reverting paramsheet to profiles

### LINKED ISSUE #472: param priority in multi-config
## Overview

This issue addresses parameter precedence in the pipeline's multi-configuration system. After extensive discussion, we've reached consensus on a **dual-mode approach** that separates single-run production use (via config profiles) from multi-run benchmarking use (via paramsheet).

---

### Two-Mode Architecture

**1. Single-Run Mode (Config Profiles)** - For production/GxP use
```bash
nextflow run nf-core/differentialabundance \
  -profile rnaseq_deseq2 \
  --limma_use_voom=false
```
- Standard Nextflow precedence: `CLI > profile > defaults`
- CLI flags override profile parameters
- Covers majority of production users

**2. Multi-Run Mode (Paramsheet)** - For benchmarking/exploration
```bash
nextflow run nf-core/differentialabundance \
  --paramsheet my_configs.yaml
```
- Paramsheet precedence: `paramsheet > CLI > defaults`
- Each row defines parameter overrides
- Runs multiple configurations in parallel
- CLI params generally don't override paramsheet (by design)
- No default paramsheet in pipeline (users provide their own)

### Key Design Principles

1. **Separation of concerns**: Profiles handle production paths, paramsheet handles benchmarking
2. **No functionality overlap**: Profile-related configurations (e.g., deseq2+rnaseq+gsea) stay in config profiles, NOT in paramsheet
3. **Minimal maintenance**: No default paramsheet to maintain
4. **Clear precedence**: Each mode has well-defined, documented precedence rules

### Timeline

- **Target**: End of March 2026
- **Goal**: Move standard configurations back to profile approach for meaningful end-to-end testing

---

### Original Issue

Paramsheet parameters had highest priority, preventing CLI overrides:

**Current (problematic):**
```
paramsheet > CLI flags / -params-file > defaults
```

**Initially desired:**
```
CLI flags / -params-file > paramsheet > defaults
```

However, achieving this in Nextflow required distinguishing CLI params from config defaults, which proved technically challenging.

---

#### ❌ Option 1: `session.cliParams` (Initially Preferred)
**Approach:** Use Nextflow's internal `session.cliParams` API
**Verdict:** Rejected - Internal API, incompatible with strict syntax (26.04+)

**Key feedback from @bentsherman:**
> The Session class is internal to the Nextflow codebase, so it isn't meant to be accessed from a Nextflow script. It won't be accessible in the strict syntax, which will be enabled by default in 26.04

#### ❌ Option 2: Schema-Based Detection
**Approach:** Compare params against `nextflow_schema.json` defaults to detect user-specified params
**Verdict:** Rejected - Edge case with bad failure mode

**Critical edge case:** If user explicitly sets param to schema default to override paramsheet, it won't be detected
- Example: schema default `param: "foo"`, paramsheet sets `"bar"`, user runs `--param=foo` → still gets `"bar"`
- Failure mode: **Silent** - user won't know override was ignored
- **Unacceptable for GxP** environments

#### ❌ Option 3: Null-Based Approach
**Approach:** Set all params to `null` in nextflow.config, handle defaults via paramsheet
**Verdict:** Rejected - Diverges too far from nf-core conventions, unclear benefit

#### ❌ Option 4: CLI Parsing
**Approach:** Parse `workflow.commandLine` with regex
**Verdict:** Rejected - Fragile, high maintenance burden

### Core Design Philosophy (from @bentsherman)

> I think there are two core assumptions in Nextflow that you will run up against:
> 1. Params define *what* is computed, config defines *how* it is computed
> 2. The pipeline script only knows the final resolved params, it does not know whether a given param came from CLI / params file / config

This insight led to reconsidering the entire approach and recognizing that trying to force CLI overrides on paramsheet violated Nextflow's design principles.

---

### Phase 1: Revert to Profile + Paramsheet Separation
1. Move standard configurations (deseq2+rnaseq+gsea, etc.) back to config profiles
2. Remove default paramsheet from pipeline
3. Document two-mode usage patterns clearly
4. Paramsheet remains available for multi-run benchmarking, but as an advanced feature

### Phase 2: Testing & Validation
1. End-to-end testing with profile-based approach
2. Validate GxP-compatible single-run mode
3. Test multi-run benchmarking scenarios

### Phase 3: Documentation
1. Clear separation of single-run vs multi-run modes
2. Document precedence rules for each mode
3. Usage examples for both patterns

---

## 🔗 Related

- Issue #465 - Initial discussion on favoring paramset over profile
- PR #623 - Implementation attempt using `session.cliParams`
- PR #624 - Follow-up implementation with filtering
- [Module interfaces (nextflow#6737)](https://github.com/nextflow-io/nextflow/pull/6737) - Future Nextflow feature that may enable cleaner solutions
- [Module params (nextflow#6769)](https://github.com/nextflow-io/nextflow/pull/6769) - Future Nextflow feature

---

## 📝 Long-Term Vision

Once Nextflow natively supports iterating over parameter configurations (via module interfaces, module params, or similar features), we may revisit unifying the two approaches. Until then, the dual-mode architecture provides a pragmatic solution that:

- ✅ Supports production/GxP use cases with standard precedence
- ✅ Enables advanced benchmarking workflows
- ✅ Maintains nf-core standards compliance
- ✅ Avoids technical edge cases and silent failures

suzannejin:
Solves #472


tests/test_affy.nf.test
```diff
@@ -39,13 +59,34 @@ nextflow_pipeline {
...
+                // Exploratory
+                exploratory_main_variable = 'contrasts'
+
+                // Differential analysis params
+                differential_max_pval        = 0.05
+                differential_max_qval        = 0.05
+                differential_min_fold_change = 1.5
+
+                // gprofiler2
+                gprofiler2_organism = 'mmusculus'
```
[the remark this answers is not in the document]
suzannejin: @copilot create a issue to check this out (indeed the GSE50790 study uses human samples)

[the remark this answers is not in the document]
suzannejin: created this issue #651 for this

pinin4fjords:
@suzannejin are you happy with this? Have you reviewed the AI's work yet?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -340,8 +340,8 @@ def methodsDescriptionText(mqc_methods_yaml) {
 // Get configurations based on whether use paramsheet or default params
 // and validate them against the schema.
 def getParamsetConfigurations() {
-    // Use paramsheet if paramset_name is provided, otherwise use default params
-    def paramsets = (params.paramset_name) ? getParamsheetConfigurations() : getDefaultConfigurations()
+    // Use paramsheet if --paramsheet is explicitly provided, otherwise use default params (profile mode)
+    def paramsets = (params.paramsheet) ? getParamsheetConfigurations() : getDefaultConfigurations()
```
grst: What's the default configuration? Why do we even have to invoke `getParamsetConfigurations()` when no paramsheet is specified?

grst (commented):
The re-added profiles and docs-related changes LGTM. 

I think currently the custom logic in `getParamsetConfigurations()` gets triggered no matter if a paramset is defined or not. I'm wondering if we could get a clearer separation of the "standard" and "custom" logic, by not even invoking that code if no paramsheet is set.

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
suzannejin: [default](https://github.com/nf-core/differentialabundance/blob/83a49a624774443be496df9cad3062542df8df06/subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf#L408) is just the `params` scope put inside the `ch_paramsheet` channel (this channel is needed to parse any other channel, see [here](https://github.com/nf-core/differentialabundance/blob/83a49a624774443be496df9cad3062542df8df06/workflows/differentialabundance.nf#L63) for example).

suzannejin:
> @suzannejin are you happy with this? Have you reviewed the AI's work yet?

Hi @pinin4fjords , it is ready for me.
Copilot's suggestion about correcting Gprofiler's database would go for a next issue imo.

nextflow.config
```diff
@@ -369,6 +369,24 @@ profiles {
         singularity.runOptions  = '--nv'
     }
 
+    // Analysis profiles (single-run mode)
+    rnaseq                    { includeConfig 'conf/rnaseq/rnaseq.config'                    }
+    rnaseq_nogtf              { includeConfig 'conf/rnaseq/rnaseq_nogtf.config'              }
```
pinin4fjords: Does this file exist?

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
pinin4fjords: Agree with @grst . The name `getParamsetConfigurations()` implies paramsheet logic, but in profile mode it just wraps `params` into a list. The calling code could handle the branching directly:

```groovy
def configurations = params.paramsheet
    ? getParamsheetConfigurations()
    : getDefaultConfigurations()
```

...with the shared validation/cleanup (the `paramsets.collect { ... }` block) extracted to something like `validateConfigurations(paramsets)`. That way the paramsheet code path is never entered when it's not needed, and the two modes are structurally separated rather than just branched inside a single function.

Not blocking, but it would make the intent clearer for future maintainers.

conf/profile/rnaseq/rnaseq.config
```diff
@@ -0,0 +1,43 @@
...
+    features_metadata_cols           = 'gene_id,gene_name,gene_biotype'
+    differential_feature_id_column   = 'gene_id'
+    differential_feature_name_column = 'gene_name'
+
+    // Observations
+    observations_id_col              = 'sample'
+    observations_name_col            = 'sample'
+
+    // Differential analysis (DESeq2)
+    differential_method              = 'deseq2'
```
pinin4fjords, line 31: We should probably document that you should now change method via profile usage, since something like this would break things due to column mismatches:

`-profile rnaseq --differential_method limma`

tests/test_rnaseq_dream.nf.test
```diff
@@ -97,12 +169,32 @@ nextflow_pipeline {
...
 
         when {
             params {
-                outdir              = "$outputDir"
-                paramsheet          = "${projectDir}/conf/test_paramsheet.yaml"
-                paramset_name       = "dream_rnaseq_decoupler"
+                outdir = "$outputDir"
+
+                // Input data (rnaseq_testdata_complex)
+                study_name        = 'SRP254919'
```
pinin4fjords: These test data items are duplicated extensively across the test files. The rnaseq test data (`SRP254919.samplesheet.csv`, `SRP254919.salmon.merged...tsv`, etc.) appears 10 times across `test_rnaseq_deseq2.nf.test` (4 cases), `test_rnaseq_limma.nf.test` (4 cases), `test_rnaseq_dream.nf.test` (1 case), and `test_nogtf.nf.test` (1 case). Before this PR these were centralized in `testdata.yaml`.

 I think we could have e.g. a `conf/testdata/rnaseq.config` with the shared test data params, which a `conf/test/rnaseq_limma.config` could include alongside the analysis config, and then `test_rnaseq_limma.nf.test` would use that as its nf-test `config`. That way each test only needs to specify its case-specific overrides, and a URL change is a one-file fix.

nextflow.config
suzannejin: @copilot remove rnaseq_nogtf

tests/test_rnaseq_dream.nf.test
suzannejin: sounds good!

suzannejin:
@pinin4fjords I have implemented what you suggested [here](https://github.com/nf-core/differentialabundance/pull/655), is this what you meant?

pinin4fjords:
> @pinin4fjords I have implemented what you suggested [here](https://github.com/nf-core/differentialabundance/pull/655), is this what you meant?

Yep, basically!

pinin4fjords (approved):
I'm going to approve this to not block you. FYI I'm on leave next week.

suzannejin:
@pinin4fjords thanks a lot :)

Issue #639, opened 2026-02-13
`--limma_confit` -> any reason not to always have it?


delfiterradas:
Remove parameter

PR #645, opened 2026-02-18
Remove `--limma_confint` parameter

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/639
- Remove `--limma_confint` parameter and always output confidence 95% intervals for logFC

## PR checklist


conf/modules.config
```diff
@@ -347,7 +346,6 @@ process {
             "--winsor.tail.p \"${meta.params.limma_winsor_tail_p}\"",
             "--p.value ${meta.params.limma_p_value}",
             "--lfc ${meta.params.limma_lfc}",
-            "--confint ${meta.params.limma_confint}",
```
grst, line 350: Right now that would remove the confidence interval, right? 
I think we should always include it. It's easy enough to filter out columns from the final dataframe if someone doesn't need it.

pinin4fjords:
Thanks for this @delfiterradas. I agree that confidence intervals should be on by default - there's no real downside to including them.

However, I'm a little uncomfortable hardcoding `--confint true` and removing the parameter entirely. The previous default of `false` matched limma's own `topTable()` default, and hardcoding a different value breaks the convention we follow elsewhere of exposing tool parameters with upstream-matching defaults.

How about we keep the `--limma_confint` parameter but just change its default to `true`? That way users get confidence intervals out of the box, but can still turn them off if they want to. It's a smaller change and stays consistent with how we handle other limma parameters.

PR #647, opened 2026-02-19
Remove `--exploratory_whisker_distance` parameter

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/642
- Remove `--exploratory_whisker_distance` parameter and set value to 1.5

## PR checklist


pinin4fjords:
Again, I don't think we should be stripping this. Is 1.5X the interquartile range always the right thing? Seems arbitrary. There should be control on that somewhere and we shouldn't be hardcoding it.

grst:
It is the de-facto standard which has been defined by Tukey in 1977 and is used by all common plotting frameworks (ggplot, plotly, seaborn, matplotlib, altair). 

Apparently in other science domains than biology, using percentiles for whisker length are sometimes uses, but that couldn't even be represented with the current parameter which is strictly factor * IQR. 

Simplicity and flexibility are competing goals, and IMO parameters that add negligible value should be removed. If there's popular demand to re-add one, I'm happy to.

pinin4fjords (changes requested):
Sorry, I don't agree that hard-coding a value is progress here.

grst:
This is not hardcoding an arbitrary value, this is not exposing an option to change the default. There's probably a hundred parameters I could set in R to customize a ggplot. Why are we exposing exactly this one? 

But anyway, I don't think this is worth arguing about.

pinin4fjords:
> This is not hardcoding an arbitrary value, this is not exposing an option to change the default. There's probably a hundred parameters I could set in R to customize a ggplot. Why are we exposing exactly this one?
> 
> But anyway, I don't think this is worth arguing about.

No, the implementation wasn't a removal of the setting, it was hard-coding it:

```
    whisker_distance = 1.5,
```

grst:
Sorry that's my bad - I have been arguing with my mental model of what the change should look like in mind without considering the actual implementation. 

I agree it shouldn't be hardcoded like that.

Issue #640, opened 2026-02-13
`--differential_foldchanges_logged` -> fold changes should always be logged.


delfiterradas:
Remove param and set to true

Issue #551, opened 2025-10-29
unification parameters lfc, pval, qval for differential expression analysis methods


suzannejin:
Currently, we have defined these parameters in multiple places: nextflow.config, profile, differential subworkflow etc.

This should be better coordinated. Also, these parameters should not be exposed to the user (it is very confusing, not even through profiles).

One solution would be to have the diff modules directly output a column names channel `tuple(meta), val(lfc_column_name), val(pval_column_name), val(qval_column_name)`

Then these parameters could be added to meta.params, and be used downstream by any procedures that require these values (eg. filtering, shinyngs app, report, etc).

--- 
- [ ] `--differential_pval_col`
- [ ] `--differential_qval_col`
- [ ] `--differential_fc_col`
- [ ] `--differential_foldchanges_logged` (see #648 for discussion and details)

grst:
So the suggestion is to remove

```
--differential_pval_col
--differential_qval_col
--differential_fc_col
```

and instead have the differential methods modules emit how their column names are called as a channel? 

Do we want to mandate that for all modules to have such an output channel? 

Other options:
 * directly add it to `meta` within the module
 * hardcode it at the pipeline level, e.g. 
   ```nextflow
   ch_deseq_res = DESEQ2.out.de_table.map{ 
        meta, it -> meta + ["p_value_col" :  "P.Value", "adj_p_value_col": "adj.P.Value", "log_fc_col" : "logFC"], it
   }
   ch_limma_res = LIMMA.out.de_table.map{ 
        meta, it -> meta + ["p_value_col" :  "pValue", "adj_p_value_col": "Padj", "log_fc_col" : "log2FC"], it 
   }
   ```

@pinin4fjords WDYT?

suzannejin:
The filtering step in the [subworkflow](https://github.com/nf-core/differentialabundance/blob/c28cb0c9c67efb676dc710b6c5a93de2249b5da6/subworkflows/nf-core/abundance_differential_filter/main.nf#L202) is already harcoding the columns/cardinality that are considered in each module. 
So I would say the most straightforward option is to let the subworkflow handle these column name related parameters too?

Then the column information can be incorporated into `meta` of the subworkflow's output channels?

grst:
works for me. Once we get clearance from @pinin4fjords we can implement this.

pinin4fjords:
Sounds OK!

grst:
As dicussed in #648, `--differential_foldchanges_logged` can be tackled the same way.

PR #648, opened 2026-02-19
Remove `--differential_foldchanges_logged` parameter

delfiterradas:
Closes https://github.com/nf-core/differentialabundance/issues/640
- Remove `--differential_foldchanges_logged` paramter and assume fold changes are always on the log scale

## PR checklist


grst:
Is there currently any module that would produce non-logged foldchanges? 

Either way, I think it falls into the same category as #551. There's no need to expose these params to the user, as its an inherent property of each method. Let's maybe wait for @pinin4fjords feedback in #551 and then address it the same way.

pinin4fjords (changes requested):
Again, this was here for a reason, it was important for some profiles to be able to define this so it worked for specific tech. IIRC it was arrays that started off logged etc, and I didn't want users to have to specify the starting log status.

You're going to need to convince me there's no fallout from this.

In general guys, things were there for a reason, as maintainers we had reasons. I would really love you to not keep stripping things assuming they're useless.

grst:
Hmm, would be good if this were documented somewhere what's the use-case. The only occurence of this param in the last release is in the schema and in the quarto report. It is not used by any profile defined in the pipeline. 

https://github.com/nf-core/differentialabundance/blob/3dd360fed0dca1780db1bdf5dce85e5258fa2253/assets/differentialabundance_report.Rmd#L388-L390


AFAIK all the currently implemented methods (limma, deseq2, dream) return logged fold changes by design (as also suggested by the column name in their results table). 

To my knowledge, limma derives the logFC from the linear model coefficients, assuming that the data is on log scale

$$
\log \text{FC} = \log \hat{X} - \log \hat{Y}
$$

where $\hat{X}$ and $\hat{Y}$ are the respective group means. 

So if the input data were not on log scale, it would rather generate a difference in means than a non-logged fold-change: 

$$
\Delta = \hat{X} - \hat{Y}
$$

Log2-transforming $\Delta$ still doesn't make it a fold-change. 

Supplying non-log-transformed data to limma may be appropriate for data that is already somewhat normally distributed. However, as stated above, in that case the `logFC` column doesn't contain a fold change at all, and log2-transforming it doesn't make it a fold-change. RMA-normalized microarray-data *is* on log-scale.

pinin4fjords:
The option was designed to support use cases where a differential method did not produce logged fold changes for reporting stages (e.g. the shiny app https://github.com/nf-core/differentialabundance/blob/3dd360fed0dca1780db1bdf5dce85e5258fa2253/conf/modules.config#L419).

Even if it's not currently used (and you're right, I don't actually see it changed in profiles right now as I'd thought), removing it makes things less flexible and could make it harder to integrate other methods in future.

Again, the parameter has no cost, and we were intentional in adding it, so I'm not in favour of stripping it. Please just ignore it in your use cases if you don't find it helpful.

grst:
> The option was designed to support use cases where a differential method did not produce logged fold changes for reporting stages

Yes, but **if** we included a method that produces non-logged fold-changes, this should be handled internally by the pipeline, and not by a user-parameter (similar to the column names discussed in #551). 

I don't want to take away the shinyngs parameter, or even the parameter in the quarto report, I just don't think this has to be exposed to the user. 

I also don't agree that parameters don't have a cost. They are a maintenance burden because with every change one has to make sure to take into account all parameters and from and end-user-perspective they add noise to the docs.

pinin4fjords:
I would 100% support a mechanism to automatically assure the correct log status based on annotation output by differential methods. Until then, I'd like to retain this control.

If you want to hide the control from the user, you can make the option hidden in the schema.

grst:
> I would 100% support a mechanism to automatically assure the correct log status based on annotation output by differential methods. 

We could tackle it together with #551? Basically adding the log-status here

https://github.com/nf-core/differentialabundance/blob/c28cb0c9c67efb676dc710b6c5a93de2249b5da6/subworkflows/nf-core/abundance_differential_filter/main.nf#L205-L222

pinin4fjords:
Oh yep, that'd work

grst:
Ok, closing this in favor of #551 then.

PR #672, opened 2026-03-23
Centralized option to set seed for all non-deterministic methods

### LINKED ISSUE #425: Centralized option to set seed for all non-deterministic methods
### Description of feature

As discussed in https://github.com/nf-core/differentialabundance/pull/420 some methods have a stochastic component and are not deterministic unless setting a random seed. 

While @pinin4fjords argues that he doesn't want to set a seed by default, it's still usefule for provenance to be able to set one. 
The most convenient way would be to set one global `--seed` parameter that then sets the seed for all appropriate methods. 
This parameter should be `null` by default and no seed set. 

Definition of done: 
 - [ ] global parameter `--seed` added and documented
 - [ ] Check that the entire output of the pipeline stays the same between two pipeline runs if a seed is set. 

e.g. this is the PCA of the same dataset in two runs with the same parameters (I don't think PCA is non-deterministic, but the VST normalization)

<table><tr><td>
Image</td><td>

Image</td></tr></table>

atrigila:
Closes #425

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- [[#665](https://github.com/nf-core/differentialabundance/pull/665)] - Add universal seed parameter ([@atrigila](https://github.com/atrigila), review by [@grst](https://github.com/grst) and [@pinin4fjords](https://github.com/pinin4fjords)).
+- [[#672](https://github.com/nf-core/differentialabundance/pull/672)] - Add universal seed parameter ([@atrigila](https://github.com/atrigila), review by [@grst](https://github.com/grst) and [@pinin4fjords](https://github.com/pinin4fjords)).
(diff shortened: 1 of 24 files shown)


grst (approved):
looks good from my end, let's wait for @pinin4fjords's approval though

CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- [[#665](https://github.com/nf-core/differentialabundance/pull/665)] - Add universal seed parameter ([@atrigila](https://github.com/atrigila), review by [@grst](https://github.com/grst) and [@pinin4fjords](https://github.com/pinin4fjords)).
```
pinin4fjords: Wrong PR ;-)

pinin4fjords (approved):
Looks good- just fix the PR number

CHANGELOG.md
atrigila:
```suggestion
- [[#672](https://github.com/nf-core/differentialabundance/pull/672)] - Add universal seed parameter ([@atrigila](https://github.com/atrigila), review by [@grst](https://github.com/grst) and [@pinin4fjords](https://github.com/pinin4fjords)).
```

atrigila: good catch!

PR #676, opened 2026-03-26
Unify de params

### LINKED ISSUE #551: unification parameters lfc, pval, qval for differential expression analysis methods

suzannejin:
Close #551 

## PR checklist


workflows/differentialabundance.nf
```diff
@@ -477,7 +477,56 @@ workflow DIFFERENTIALABUNDANCE {
     // Also note that these channels, the meta contain other info apart from the base paramset meta.
     // Hence we create a key using the simple paramset meta with only meta.id, meta.paramset_name and meta.params,
     // by setting 'use_meta_key' to true. This will facilitate later on to join/combine channels.
-    ch_differential_results = prepareModuleOutput(ABUNDANCE_DIFFERENTIAL_FILTER.out.results_genewise, ch_paramsets, meta_keys_to_remove=['differential_method'], use_meta_key=true) // key, meta, results
+    // Define method_params at workflow level (one time definition)
+    def method_params = [
```
suzannejin: I'll favour add the method_params manipulation at the very beginning of the pipline, inside `subworkflows/local/utils_nfcore_differentialabundance/main.nf`.
Directly as part of the paramset parsing process.

workflows/differentialabundance.nf
```diff
@@ -705,6 +706,26 @@ workflow DIFFERENTIALABUNDANCE {
...
     // Create a channel with the differential results and the corresponding map with
     // the contrast entries
+    def differential_runtime_param_keys = [
+        'differential_fc_column',
+        'differential_pval_column',
+        'differential_qval_column',
+        'differential_foldchanges_logged'
+    ]
+
+    ch_differential_runtime_params = ch_differential_results
```
suzannejin: I kind of feel having to handle these parameters separately as runtime params can be quite confusing and error prone. 
What if you provide these parameters already at the very very beginning of the pipeline, at the step of creating `ch_paramset`.
Then the `ch_paramset` will have them in meta.params since the very beginning.
And the `abundance_differential_filter` subworkflow can check if these values are already provided, and if not, add them.

In this way, when you deal with meta.params you wont need to create so many special cases, conditions, and mappings.

antoniasaracco:
Hi @suzannejin @pinin4fjords, thanks a lot for your feedback it helped a lot! I have implemented your comments and it's ready for review. Please let me know what you think. Thanks!

subworkflows/nf-core/abundance_differential_filter/main.nf
```diff
@@ -18,6 +18,41 @@ def mergeMaps(meta, meta2){
     }
 }
 
+def getDifferentialMethodParams(differential_method) {
```
pinin4fjords, line 21: **Severity: LOW — role-clarifying comment**

The two method→column maps in this PR (this one, and `getDifferentialMethodRuntimeParams` in `subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf:170`) were the agreed outcome of the discussion on nf-core/modules#11221: the subworkflow owns FILTER columns, the pipeline owns DISPLAY columns, neither reaches into the other's domain.

The propd row makes the role split visible: here `stat_column='significant'` (the boolean filter column the module emits), there `differential_qval_column='rcDdis'` (the continuous score the report displays). Different columns, different jobs.

Worth a one-liner so a future maintainer doesn't try to "unify" them and break propd filtering:

```suggestion
// Per-method column mapping for FILTERING. The pipeline-side report/shiny
// columns are stamped into meta.params separately by
// utils_nfcore_differentialabundance_pipeline::getDifferentialMethodRuntimeParams;
// for propd these are deliberately distinct (`significant` here vs `rcDdis`
// in the pipeline map).
def getDifferentialMethodParams(differential_method) {
```

subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf
```diff
@@ -166,6 +167,45 @@ workflow PIPELINE_COMPLETION {
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 */
 //
+def getDifferentialMethodRuntimeParams(differential_method) {
```
pinin4fjords, line 175: **Severity: LOW — mirror comment**

Same as the note in the subworkflow side. Worth recording the role distinction in code so the propd `pval_column == qval_column == rcDdis` choice (raised but not directly answered on nf-core/modules#11221) is explicit rather than living only in PR threads.

Also a question for @suzannejin / @antoniasaracco that's still open from the modules PR: is propd's `differential_pval_column` and `differential_qval_column` both being `rcDdis` intentional? For deseq2/limma/dream they are distinct columns; for propd they collapse to the same continuous score. If that's the right semantics for the report (display the same value in both pval and qval slots), worth saying so in the comment below.

```suggestion
// Per-method DISPLAY columns stamped into meta.params at paramset-parse
// time, used downstream by report.qmd, shinyngs and the output DSL.
// FILTER columns are owned separately by the abundance_differential_filter
// subworkflow's getDifferentialMethodParams (deliberately different for
// propd: `significant` is the filter column, `rcDdis` is the display score).
def getDifferentialMethodRuntimeParams(differential_method) {
```

pinin4fjords (commented):
AI-assisted review (Claude, on behalf of @pinin4fjords).

Thanks for picking up #551 — the architecture matches what was settled on nf-core/modules#11221: subworkflow keeps only the filter columns it actually uses, pipeline-side `addDifferentialRuntimeParams` stamps the display columns into `meta.params` at paramset-parse time. That's the right separation; the "two maps" pattern is the deliberate outcome of it, not duplication to clean up.

---

**HIGH — incomplete profile cleanup**

The PR removes `differential_fc_column` / `_pval_column` / `_qval_column` from `affy.config`, `generic_matrix.config`, `rnaseq.config`, and the test paramsheet, but leaves them in:

- `conf/profile/maxquant/maxquant.config:35-37`
- `conf/profile/soft/soft.config:38-40`
- `conf/profile/rnaseq/rnaseq_limma.config:26-28`
- `conf/profile/rnaseq/rnaseq_dream.config:26-28`
- `conf/profile/generic_matrix/generic_matrix_dream.config:21-23`

`addDifferentialRuntimeParams` only fills *missing* keys (`!containsKey`), so the leftovers act as profile-level overrides rather than breaking anything. But they defeat the PR's stated goal (per #551, these params shouldn't be expressed in user-editable config at all), and the choice of which profiles were cleaned vs left alone looks arbitrary. If the runtime mapping ever changes — e.g. limma adopts a new column convention — these five profiles will silently keep the old values.

Recommend deleting the three lines from each of the five files above in this PR. (I couldn't post inline suggestions because none of those files appear in the diff.)

---

**LOW — PR description and CHANGELOG**

- The PR body is the unfilled template. Worth a few lines describing what moved out of user-facing config, what now derives the values automatically, and that profile/paramsheet overrides still take precedence (because `addDifferentialRuntimeParams` only fills missing keys).
- No CHANGELOG entry for #676 yet — should land under v1.6.0dev `### Removed` (for the four removed schema params) and/or `### Changed`.

---

Two inline comments below are LOW-severity role-clarifying notes on the two `getDifferentialMethod*Params` functions.

pinin4fjords (approved):
I think we might have to re-examine this stuff at some point, approving for now!

Issue #657, opened 2026-03-06
rounding is turned on by default in the dev branch


akaviaLab:
### Description of the bug

In the dev branch, rounding is turned on by default
https://github.com/nf-core/differentialabundance/blob/787c8e93e44a476ecc4d4db20903102cc86f03c7/nextflow.config#L27

The way the rounding scripts work, since they use round, it will turn eveything below the round_digits to 0, which removes a lot of data. 

1. This should probably be reverted, so that rounding happens only in test, as per conversation with @pinin4fjords on slack.
2. I would suggest (feature enhancement) to change the rounding code to something like `apply(data.frame(comp.results), 2, function(x) sprintf(paste0("%." opt\$round_digits, "g"), x))` throughout the modules that round down.
This way, rounding will avoid floating point errors when testing, and if someone turns it on when running, it will not destroy data. I'd be happy to tackle this if the idea of changing it is approved.

I'm harping on this, because I believe Seqera has rounding on by default, and then I was bitten by having lots of zeros. I also think that the average user might think "Sure, I'd love to round data" and then be confused why there are so many zeros.
If you guys want to keep the rounding as is, can you please modify the documentation to warn people against rounding?

### Command used and terminal output

```console

```

grst:
Closed via #681

PR #681, opened 2026-03-27
Change `round_digits` default

atrigila:
For: https://github.com/nf-core/differentialabundance/issues/657

- Rounding now only happens in the tests, not by default.

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -24,7 +24,7 @@ params {
 
     // Output options
     outdir                     = null
-    round_digits               = 4
+    round_digits               = -1
 
     // Paramsheet options
     paramset_name              = null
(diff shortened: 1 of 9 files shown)


nextflow.config
```diff
@@ -24,7 +24,7 @@ params {
 
     // Output options
     outdir                     = null
-    round_digits               = 4
+    round_digits               = -1
```
grst, line 27: would null be a more intuitive/consistent default? Or is there any reason why this wouldn't work?

atrigila: It is the same, as per the schema: `"To how many digits should numeric output in different modules be rounded? If -1 or null, will not round."`.

Do you prefer it to be changed to `null`?

grst: ok, fine for me to leave as is then.

pinin4fjords (approved):
Makes sense - rounding by default is a footgun for small values. Thanks @atrigila.

Issue #666, opened 2026-03-20
Specifying model with intercept leads to incorrect result


grst:
### Description of the bug

When I specify a model with intercept, e.g. 

`~ BORC_simple`

at some point in the pipeline an intercept gets added automatically: 

<img width="1175" height="397" alt="Image" src="https://github.com/user-attachments/assets/100bdaab-b5be-49ad-9076-ceacaed5670d" />

Doing that implicitly produces unexpected results. 

For complex designs, it's quite common to deliberately set a base factor level and then contrasts will be relative to that level. This simplifies the specification of some contrasts. 

---

#### command

```bash
nextflow run delfiterradas/differentialabundance -r 67d8916 \
    -params-file src/params.yml \
    -profile rnaseq,bi,cluster \
    -c src/nextflow.config \
    -resume
```

#### params.yml

```yaml
study_name: "baseline_response_deseq2"

features_id_col: "gene_id"
features_name_col: "GENESMBL"

input: "<path redacted>/input.tsv"
matrix: "<path redacted>/matrix.tsv"
features: "<path redacted>/features.tsv"
contrasts: "src/contrasts.yml"
outdir: "output"

filtering_min_samples: 5
filtering_grouping_var: "BORC_simple"

shinyngs_build_app: false
```

#### contrasts.yml

```yaml
contrasts:
  - id: BORC_simple_SD_vs_CRPR_formula_intercept
    formula: ~ BORC_simple
    make_contrasts_str: "BORC_simpleSD"
# should be equivalent to (because CR/PR is alphabetically first and therefore becomes the baseline level)
 - id: BORC_simple_SD_vs_CRPR
   comparison: ["BORC_simple", "CR/PR", "SD"]
```

@delfiterradas @antoniasaracco @atrigila you should be able to reproduce this with the mock data you already have. 

### Command used and terminal output

```console

```

grst:
This only appears to happen with DESeq2, at least with dream it worked fine.

atrigila:
I think this is because of this line:
https://github.com/nf-core/differentialabundance/blob/339f0358b9bf4cf9bc86884ba37abd587e0c9d00/modules/nf-core/deseq2/differential/templates/deseq2_differential.R#L355

PR #684, opened 2026-03-30
Fix intercept in `deseq2`

atrigila:
Fixes https://github.com/nf-core/differentialabundance/issues/666


grst:
The actual change is uncontroversial. Do you have any clue where all these changes in the test files come from?

atrigila:
The order of execution changed when I run the tests and took the snapshots, but the results are the same.

grst (approved):

PR #686, opened 2026-04-01
Improve abundance type docs and schema description

### LINKED ISSUE #667: study_abundance_type parameter is confusing
### Description of feature

`study_abundance_type` is prominently advertised in the docs in the "base input options" are "required". It currently only supports "counts" and "intensities". 

As far as I can tell, the only implication of `study_abundance_type` is whether the headline in the quarto report reads "counts" or "intensities": https://github.com/nf-core/differentialabundance/blob/339f0358b9bf4cf9bc86884ba37abd587e0c9d00/assets/differentialabundance_report.qmd#L1238

Is this worth an extra parameter, and couldn't this derived from something else?

--- 

The main problem I have with this paramters is that I would have assumed it has some implications on the analysis: e.g. automatically choosing the right methods params depending on whether I input counts or continuous values. However, this is not the case. It is completely up to me to choose an appropriate method/profile that can deal with the respective input type (And I think it's fine, adding appropriate profiles for count/non-count data is the way to go IMO). 

The second problem I have is that there exists more than "counts" and "intensities". I might have pre-transformed gene expression data, cell-type fractions, or other continuous values that are neither counts nor intensities but can very well be analyzed using a generic method such as limma. 

---

@pinin4fjords do you have thoughts on this?

atrigila:
Closes #667 
- Make `study_abundance_type` a hidden parameter and do not make it a required parameter to run the pipeline.


nextflow_schema.json
```diff
@@ -56,9 +56,10 @@
                 "study_abundance_type": {
                     "type": "string",
                     "default": "counts",
-                    "description": "Type of abundance measure used, platform-dependent.",
-                    "enum": ["counts", "intensities"],
-                    "fa_icon": "fas fa-keyboard"
+                    "description": "Label describing the abundance values in the report heading.",
+                    "help_text": "Optional display label used in report text (for example 'counts', 'intensities', or any custom term). This does not alter statistical method selection.",
```
grst:
```suggestion
                    "help_text": "Optional display label used in report text (for example 'counts', 'intensities', or any custom term). This does not alter statistical method selection. Typically set internally when choosing a profile.",
```

assets/differentialabundance_report.qmd
```diff
@@ -1260,7 +1260,7 @@ method_description <- switch(
 cat(method_description)
 ```
 
-### Differential `{r} params$meta$params$features_type` `{r} params$meta$params$study_abundance_type`
+### Differential `{r} params$meta$params$features_type` `{r} ifelse(is.null(params$meta$params$study_abundance_type) || params$meta$params$study_abundance_type == "", "abundance", params$meta$params$study_abundance_type)}`
```
grst: Maybe just set the default to `"abundance"` and expect it to be always there?

pinin4fjords (approved):

PR #662, opened 2026-03-11
migrate to new workflow outputs

suzannejin:
This solves #611 (linked to parent issue #471), by 1) migrating to the new workflow output format and 2) remove paramset_names when running modules.

## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core lint`).
- [ ] Ensure the test suite passes (`nf-test test main.nf.test -profile test,docker`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

PR #687, opened 2026-04-07
Emit individual channels from inner workflow

pinin4fjords:
## Summary

- Refactors the inner workflow (`DIFFERENTIALABUNDANCE`) to emit each output as its own clean, typed channel (~62 individual emits) instead of mixing them into 7 broad category channels with name-tags
- Moves all name-tagging (`['name', meta, file]`) and category mixing logic into `main.nf`'s entry workflow, keeping publishing concerns in one place
- The `output {}` block is unchanged - it still consumes the same 7 category channels

This follows the principle established in nf-core/modules#11024: components should not mix disparate types into single channels or make assumptions about how consuming workflows will publish. The inner workflow now emits clean channels that are straightforward to consume, and will port more easily to future Nextflow enhancements (record types, etc.).

## Changes

- **`workflows/differentialabundance.nf`**: Removed `ch_preprocessing`, `ch_differential`, `ch_functional`, `ch_plotting`, `ch_shinyngs`, `ch_report` category channels and all `.mix()` chains. Each output is now emitted individually.
- **`main.nf`**: `NFCORE_DIFFERENTIALABUNDANCE` now passes through all ~62 individual channels. The entry workflow builds the 7 category channels with name-tagging in its `main:` block before publishing.

## Test plan

- [ ] CI tests pass (same test suite as PR #662)
- [ ] Output directory structure is unchanged (same `output {}` block)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -52,13 +52,95 @@ workflow NFCORE_DIFFERENTIALABUNDANCE {
     )
 
     emit:
-    preprocessing = DIFFERENTIALABUNDANCE.out.preprocessing
-    differential  = DIFFERENTIALABUNDANCE.out.differential
-    functional    = DIFFERENTIALABUNDANCE.out.functional
-    plotting      = DIFFERENTIALABUNDANCE.out.plotting
-    shinyngs      = DIFFERENTIALABUNDANCE.out.shinyngs
-    report        = DIFFERENTIALABUNDANCE.out.report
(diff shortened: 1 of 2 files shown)


suzannejin:
@pinin4fjords thanks a lot, this looks good!

PR #692, opened 2026-04-09
Add profiles for generic matrix

grst:
This PR adds profiles for generic, pre-transformed matrices (typically non-count data) that can be used with either limma or dream.


atrigila (approved):
This looks good to me. I will follow with a test case for this profile in PR #694

atrigila:
(merging in order to be able to continue in #694)

PR #705, opened 2026-04-27
Update modules to use seqera containers

### LINKED ISSUE #697: Fully switch to seqera containers
### Description of feature

Currently we have a mix between biocontainers and seqera containers. 
Since especially deplot.singularityproject tends to make problems from time to time, it would be nice to migrate the remaining modules to use seqera containers.

atrigila:
Update modules to use seqera containers and update all modules and subworkflows. 

The `ZIP` / `MAKE_REPORT_BUNDLE` module had to be patched in order not to emit topics, which was causing the whole pipeline to lag and avoiding it to end. 

Closes #697


modules.json
```diff
@@ -7,146 +7,147 @@
...
                     "proteus/readproteingroups": {
                         "branch": "master",
-                        "git_sha": "41dfa3f7c0ffabb96a6a813fe321c6d1cc5b6e46",
+                        "git_sha": "07cee52e08af8100eee4ea1b53abf0660df77701",
                         "installed_by": ["modules"]
                     },
                     "quartonotebook": {
                         "branch": "master",
-                        "git_sha": "e753770db613ce014b3c4bc94f6cba443427b726",
+                        "git_sha": "e753770db613ce014b3c4bc94f6cba443427b72",
```
pinin4fjords: Truncated SHA — 39 chars instead of 40, missing the trailing `6`. Introduced in commit `98b1c88c` while re-pinning quartonotebook back to its dev-branch SHA. Happens to be a valid prefix so it may resolve in practice, but `nf-core lint` expects full 40-char SHAs.

```suggestion
                        "git_sha": "e753770db613ce014b3c4bc94f6cba443427b726",
```

pinin4fjords (commented):
Mostly LGTM after a read-through of `modules.json`, both subworkflow `main.nf`s, the workflow change, the `zip.diff` patch, the `decoupler.py`/`gprofiler2_gost.R` templates, and a skim of the remaining module `main.nf` diffs.

One real gremlin (inline) and a couple of low-severity notes:

- The `quartonotebook` `git_sha` got truncated by one character in commit `98b1c88c` ("fix local copy of module does not match remote") — see inline suggestion. The SHA is 39 chars instead of 40; happens to be a valid prefix of the intended SHA but strict lint will flag it.
- `modules/nf-core/gunzip/main.nf` has mixed tabs/spaces in the new `script:`/`stub:` blocks (lines starting `def extension = ...` / `def name = ...`). Confirmed this is identical to upstream `nf-core/modules` at the pinned SHA, so leaving as-is is correct.
- The `zip.diff` patch evaluated cleanly: only removes `, topic: versions` from the version-emit line, preserves `emit: versions_zip`, and the only pipeline consumer of `MAKE_REPORT_BUNDLE` reads `.zipped_archive`, so no downstream consumers break. Good fix for the topic-aggregator-can't-also-emit-to-the-topic constraint.

pinin4fjords (approved):
Approving with that minor point

modules.json
grst: good catch, but how would that happen? I doubt the hashes were updated by hand?

PR #726, opened 2026-05-13
fix(schema): model comma-separated report_file in pattern

### LINKED ISSUE #725: Schema for `report_file` doesn't model the comma-separated multi-report mode
## Background

`nextflow_schema.json:1129` declares the `report_file` pattern as:

```json
"pattern": "^\\S+\\.(Rmd|qmd|ipynb)$"
```

But `workflows/differentialabundance.nf:864-870` splits `meta.params.report_file` on `,` to support rendering multiple reports per paramset, and `tests/default.nf.test:69-99` exercises exactly that path:

```nextflow
report_file = "$projectDir/tests/quarto.qmd,$projectDir/assets/differentialabundance_report.qmd"
```

The pattern *accidentally* matches comma-separated values because `\S+` is greedy enough to swallow the comma and the next path before reaching the trailing `\.(qmd|...)$`. Verified empirically on Nextflow 26.04.1:

```groovy
def regex = ~/^\S+\.(Rmd|qmd|ipynb)$/
'a.qmd,b.qmd' ==~ regex                      // true
'a.qmd,b.qmd,c.qmd' ==~ regex                // true
'/abs/path/a.qmd,/abs/path/b.qmd' ==~ regex  // true
'noextension' ==~ regex                       // false
```

So today this works but only because of a regex coincidence, not because the schema models the feature. A user-friendly fix would be either:

1. Relax the pattern to explicitly allow comma-separated paths (and validate each segment), or
2. Expose the field as `array | string` in the schema and let the comma-split happen in code.

Both would also make `--help` / `nf-schema`-generated docs reflect the actual supported usage.

Originally surfaced during the dev-vs-master review for 1.6.0. Related to #712 / #716 (now closed/merged) which covered other schema-completeness issues.

pinin4fjords:
Closes #725.

## Summary

- `report_file` is documented as accepting a single Quarto/Rmd/ipynb path, but `workflows/differentialabundance.nf:864-870` splits the value on `,` to render multiple reports per paramset (and `tests/default.nf.test` exercises that path). The previous schema pattern `^\S+\.(Rmd|qmd|ipynb)$` only accidentally matched those inputs because `\S+` is greedy enough to swallow commas, and it would accept mixed-extension nonsense like `foo.txt,bar.qmd`.
- Tighten the pattern so each path segment uses `[^,\s]+` and the comma-separated form is explicit: `^[^,\s]+\.(Rmd|qmd|ipynb)(\s*,\s*[^,\s]+\.(Rmd|qmd|ipynb))*$`. Each segment's extension is now actually validated.
- Update `description`/`help_text` so `--help` and the nf-schema-generated docs reflect the multi-report usage.

## Test plan

- [x] `nf-core pipelines schema lint nextflow_schema.json` clean.
- [x] Regex unit cases: accepts single path, comma-separated paths (with/without spaces) and absolute paths; rejects `noextension`, `a.txt`, `a.txt,b.qmd`, `a.qmd,b.txt`, trailing/leading commas, empty string, whitespace-within-path.
- [x] `nf-test test tests/default.nf.test@3142bf2f --profile docker` (the "multiple reports - csv contrasts" case) passes against the new schema.
- [x] Sanity-checked that an invalid mixed-extension value (`--report_file 'foo.txt,bar.qmd'`) now triggers an nf-schema validation error, where the old pattern accepted it.


atrigila (approved):
Looks go

Issue #541, opened 2025-10-23
update nf-core/modules propr_propd and make it work in the pipeline


suzannejin:
- Update propr module
- Add paramset option for running it
- Add tests in the pipeline to run it
- Modify reports/other details accordingly to make it usable

PR #727, opened 2026-05-13
feat: enable propd differential method and grea functional method

### LINKED ISSUE #724: Decide on `propd` / `grea`: enable in schema or remove from subworkflows
## Background

The `abundance_differential_filter` and `differential_functional_enrichment` subworkflows (from nf-core/modules) include branches for `propr/propd` and `propr/grea`, but the pipeline schema doesn't expose those as user-selectable values:

- `subworkflows/nf-core/abundance_differential_filter/main.nf:9,146-149` runs `PROPR_PROPD` when `meta.differential_method == 'propd'`.
- `subworkflows/nf-core/differential_functional_enrichment/main.nf:10,141-144` runs `PROPR_GREA` when `meta.functional_method == 'grea'`.

But:
- `nextflow_schema.json:491` — `differential_method.enum` is `[\"deseq2\", \"limma\", \"dream\"]`.
- `nextflow_schema.json:844` — `functional_method.enum` is `[\"null\", \"gsea\", \"gprofiler2\", \"decoupler\"]`.

Confirmed on Nextflow 26.04.1 against `origin/dev` `f9a174b5`:

```
\$ nextflow run . -profile test,docker --differential_method propd ...
ERROR ~ /differential_method (propd): Expected any of [[deseq2, limma, dream]]
```

So today the propd/grea code paths exist in the subworkflows but are unreachable through the pipeline's parameter interface.

## What to decide

Whether we want to plumb them through. The subworkflows themselves live in nf-core/modules and don't need any change either way — this is just a question of whether the differentialabundance pipeline exposes them.

- **Plumb them through**: add `propd` to `differential_method.enum` and `grea` to `functional_method.enum`, add the relevant `--propd_*` / `--grea_*` params and modules.config entries, document in `docs/usage.md`, and refresh the metro map (#710). #541 already tracks the `propd` half.
- **Leave them dormant**: no schema change, no docs change. The subworkflow branches stay where they are; we just don't advertise them.

No rush — flagging here so it can ride along with whichever direction makes sense for 1.6.0 or later. Relates to #541.

pinin4fjords:
## Summary

- Add `propd` to `differential_method.enum` and `grea` to `functional_method.enum`, with corresponding `propd_*` / `grea_*` schema entries and `nextflow.config` defaults
- Add `withName: PROPR_PROPD` and `withName: PROPR_GREA` blocks in `conf/modules.config`
- Route propd's adjacency matrix into the functional input channel when `functional_method == 'grea'`
- Add `rnaseq_propd` and `rnaseq_propd_grea` analysis profiles, a `rnaseq_propd_grea` paramsheet entry, and a `test_rnaseq_propd_grea` test profile
- Add `tests/test_rnaseq_propd_grea.nf.test` + snapshot, validated end-to-end on Nextflow 25.10.3 / Docker

## Closes

- #724
- #541 (partial - propd/grea results are written to disk and surface in the parameters table, but the per-method rendering branch in `differentialabundance_report.qmd` is still gsea/gprofiler2-only; tracked in #728)

## Notes

- The propd column mapping (`LFC` as the effect-size column, `rcDdis` as both the display p-value and q-value, `significant` as the filter column) is stamped onto each paramset at parse time by `getDifferentialMethodRuntimeParams()` in `subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf` (existing runtime-params infrastructure), so the new paramset YAML doesn't need to set those columns explicitly.
- propd is normalization-free and does not emit a normalised matrix; the new profile only requests the `raw` assay. The report-input join needed `remainder: true` on `ch_differential_model` so methods without a model file (propd) still reach the report.
- The `rnaseq_propd` profile sets `differential_min_fold_change = 1.5` rather than inheriting the 2.0 global default, because propd's LFC is a per-gene log-ratio mean (not a condition-vs-condition fold change) and the 2.0 cutoff is correspondingly strict.
- grea consumes propd's adjacency matrix and so cannot be paired with any other differential method - this is enforced by validation in `utils_nfcore_differentialabundance_pipeline`.

## Test plan

- [x] `nf-test test tests/test_rnaseq_propd_grea.nf.test --profile docker` passes (validated locally on Nextflow 25.10.3)
- [x] Schema lint clean
- [x] `--differential_method propd` no longer rejected by the schema
- [x] Quarto HTML report renders for the propd + grea paramset


atrigila (approved):
I tested the new profile and it worked. I cannot comment on the method itself / results as I haven't used this before.

pinin4fjords:
Thanks @atrigila ! Me neither, I'm relying on the intent of @suzannejin and co in initiating these changes before.

suzannejin:
Hey @pinin4fjords and @atrigila , really really thank you for all the work!!!! I appreciate it a lot.
The propd and grea modules were not updated yet because we were still waiting for some final polishing of the R package itself, that is why they were on hold for properly integrating into the pipeline as new changes may get introduced soon.
Yet thanks a lot for this merge as it will help @caraiz2001 , who is in charge now, so she will only need to update the modules but not the pipeline logic itself :)

Issue #599, opened 2026-01-08
Refactor quarto report


grst:
### Description of feature

The quarto report has grown quite a bit and different parts are redundant and/or inconsistent. The whole report should be refactored holistically:

 * Have one data wrangling section at the top, and reuse dataframes throughout the report
 * use dplyr/tidyr for all dataframe manipulations
 * use plotly wherever possible
 * Replace as many of the `cat(paste0("### heading"))` with plain markdown
 * consider having a file with helper functions

Maybe the report could also be split up into independent "modules" that can be activated via params, in case someone wants to disable certain sections, see e.g. https://github.com/nf-core/differentialabundance/pull/487

pinin4fjords:
PR #729 lands most of the asks here (modularisation via `disable_report_modules`, tidyverse-only df manipulation, helper consolidation, lots of `cat()` heading removals). Two more small follow-ups went in just now in 1ece27c3:

- biotype bar plot now goes through `plotly::ggplotly()` like the rest of the per-contrast widgets
- the static `### Upset plot` heading is no longer emitted once per p-value type

One thing worth recording before we close this out: the original ask included "use plotly wherever possible", and I had a go at applying that to the exploratory sample-level boxplot (`ggplot_boxplot` -> `plotly::ggplotly(...)`). It works on the test data, but it's not safe in general.

`ggplotly()` over a `geom_boxplot` (and `shinyngs::plotly_boxplot` for the same reason) ships every raw per-gene value to the browser instead of pre-computing the 5-number summary. That makes the trace payload scale as N_genes * N_samples. Concrete numbers from the test profile:

- box-plots section: 101 KB (static PNG) -> 347 KB (plotly), ~3.4x growth for ~3K filtered genes x ~20 samples
- extrapolating to a typical real RNA-seq study (60K genes x 30 samples x 3 assay matrices) that one section alone gets to ~50-80 MB

shinyngs's own `plotly_boxplot` carries this exact warning in its roxygen: "NOT CURRENTLY USED DUE TO RESOURCE REQUIREMENTS ON LARGE MATRICES".

UX-wise the trade is also poor: a boxplot already abstracts the data into a summary, so hover-on-every-point isn't adding information, just bytes.

So the principle to apply (and worth keeping in mind for future plot additions) is something like:

- aggregated/summary data (biotype counts, PCA scores, scree, volcano-style scatters) -> plotly is a clear win
[...]

Issue #718, opened 2026-05-12
Paramsheet validation prints "Error: null" before the real error


pinin4fjords:
# Paramsheet validation failure prints "Error: null" before the real error

## Background

When pipeline validation fails (e.g. invalid input path, malformed contrast file), the first line of output is unhelpful:

```
Validation failed for paramsheet row: rnaseq. Error: null
/input (https://...samplesheet.csv): the file or directory ... does not exist
/contrasts (/tmp/bad_contrast.csv): Validation of file failed:
    -> Entry 1: Value is [string] but should be [object]
    -> ...
```

The first "Error: null" line is misleading — users see "null" and may think the framework hit an internal bug. The actually-useful nf-schema validation errors are printed underneath, but the leading "Error: null" creates noise on every validation failure.

## Root cause

`subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf:368-371`:

```groovy
} catch (e) {
    // if validation fails, print error message and exit
    println("Validation failed for paramsheet row: ${paramset.paramset_name}. Error: ${e.message}")
    validate(notnullparams, "${projectDir}/nextflow_schema.json")
}
```

`e.message` is `null` for the nf-schema validation exception thrown by `validate()`. The intent appears to be: print a brief banner, then re-call `validate()` so it raises and prints its own detailed error. But the banner ends up showing "Error: null" because the original exception's message is empty.

### 1. Either print a meaningful banner or drop it entirely

Option A — drop the leading banner and let nf-schema produce its own error:

```diff
 } catch (e) {
-    // if validation fails, print error message and exit
-    println("Validation failed for paramsheet row: ${paramset.paramset_name}. Error: ${e.message}")
+    // Surface the paramset name; nf-schema will then produce a detailed error.
+    log.error "Validation failed for paramsheet row: ${paramset.paramset_name}"
     validate(notnullparams, "${projectDir}/nextflow_schema.json")
 }
```

Option B — extract the real message from the exception:

```diff
-    println("Validation failed for paramsheet row: ${paramset.paramset_name}. Error: ${e.message}")
+    def detail = e.message ?: e.cause?.message ?: e.class.simpleName
+    log.error "Validation failed for paramsheet row: ${paramset.paramset_name}: ${detail}"
```

Either way, the goal is: no user-visible "Error: null".

### 2. Verify nothing relies on the literal string "Error: null"

Grep tests and snapshot files for "Error: null" to confirm no snapshot is matching that exact string.

```
[...]

PR #729, opened 2026-05-13
refactor(report): consolidate helpers and modernise the default qmd

pinin4fjords:
## Summary

The default Quarto report (`assets/differentialabundance_report.qmd`) had grown to ~1950 lines of intermingled helper functions, monolithic chunks and prose, which made the file painful to wrangle. This PR consolidates the implementation into a single helpers chunk near the top so the body becomes a short, navigable outline of `render_*` calls. Three small intentional output improvements are called out below; everything else is byte-for-byte equivalent rendered output.

**Self-contained**: no new files, no workflow plumbing, no new environment dependencies. Everything lives in the single qmd, exactly as before.

## What changed (structure)

- **Single helpers chunk** (`#| include: false`) holds all utility, plot, and section-renderer functions. The prose body below is now pure orchestration: `render_filtering_summary()`, `render_pca_panels()`, `render_de_details(i)`, `render_gsea_section()`, etc.
- **`p <- params$meta$params` alias** replaces ~147 nested `params$meta$params$X` references.
- **`captioned_table()`** unifies the reactable + `tagList` + caption-div pattern (14 hand-rolled instances → one helper).
- **`build_de_combined()`** separates the per-contrast Direction-annotated frame construction from `render_de_table`, so the DE table and boxplots share one source of truth.
- **`blank(x)`** helper replaces the repeated `is.na(x) || x == "" || is.null(x)` triplet.
- **`name_contrast`** is now row-shaped (takes a list) and called via `pmap_chr` over the contrasts frame instead of scalar-per-row indexing.

## What changed (style)

- Native `|>` everywhere (no `.` placeholders used)
- `lapply`/`vapply`/`sapply` → `purrr::map_*`/`walk`/`iwalk`/`pwalk`/`walk2`
- Anonymous `function(x)` → `\(x)`
- All `for` loops with side-effect bodies → `walk`/`iwalk`/`pwalk` (state-accumulating loops kept as `for`)
- `setNames` → `set_names`; `names(x) <- foo` after a map folded into `set_names(map(...), foo)`
- `head(x, 1)` / `tail(x, -1)` → `first(x)` / `x[-1]`; nested `gsub` → `str_remove_all`
- Tidyverse-meta-package namespacing dropped (`dplyr::`, `tidyr::`, `ggplot2::`, `purrr::`, `tibble::`, `readr::`, `stringr::`) since `library(tidyverse)` is loaded. Non-tidyverse namespacing kept (`plotly::`, `htmltools::`, `reactable::`, `ComplexHeatmap::`, ...)
- gprofiler2 section: `data.frame(..., check.names=FALSE)` → `transmute` with backticked names; `read.table` → `read_tsv` for consistency with the rest of the file
- `apply(contrasts, 1, paste, ...)` (for default id) → `pmap_chr` to avoid frame-to-character-matrix coercion
- Plotly PCA marker dispatch: dropped the `do.call(function(...) plotly::add_markers(fig, ...), ...)` anonymous-closure wrapper in favour of `do.call(plotly::add_markers, c(list(fig), common, ...))`

## Intentional behaviour / output changes

Three things that change rendered output deliberately. All were pre-existing bugs in dev:

1. **NA-coercion fix** in the observations-metadata "useful columns" detection. Dev used `apply(observations, 2, function(x) max(nchar(x)))`; `nchar(NA)` is NA, `max(NA, ...) = NA`, `NA <= 20` is filtered out by `which`. Any column with even one NA was silently dropped. Replaced with a per-column `map_int` using `na.rm = TRUE`. For the test dataset there are no NAs in the candidate columns so the rendered output is unchanged; on real datasets with sparse metadata the metadata table will now include columns that were previously dropped.

2. **Filtering-stats chunk rendered cat() text as a code block**. The dev chunk lacked `results='asis'`, so `cat("The following tables summarize...\n\n")` was wrapped in `<pre><code>...</code></pre>` like code output. Setting the chunk to `results='asis'` (required for the refactor's print-from-function pattern) also turns this into a proper paragraph.

3. **Typo fix**: `assocations` → `associations` in the PCA-vs-metadata heatmap prose.

## Verification

End-to-end on `nf-dev-differentialabundance` with `nextflow run -profile test,docker` against `NXF_VER=25.10.3` - pipeline completes successfully.

**Comprehensive widget/output audit** comparing rendered HTML against dev's:
- Widget counts (reactable / plotly / d3heatmap / img): identical
- Widget inline styles and dimensions: identical
- Section count: 45 = 45
- Headings: identical
- Body prose paragraphs: identical apart from the 2 intentional fixes above
- Embedded JS bundle warnings: identical count (these come from plotly.min.js, not the report)
- Final ZIP sizes: 2,490,276 vs 2,491,046 bytes (delta = unique widget IDs)

## Notes for future qmd contributors

A few subtle gotchas surfaced during the refactor - worth documenting:
- **Invisible returns**: if a function (yours or upstream) ends with an assignment, the result is invisible and R's chunk auto-print will skip it. Use bare-name-on-next-line or wrap in `print()`.
- **htmlwidget print dispatch**: don't `print()` a raw htmlwidget from inside a helper - `print.htmlwidget` is the interactive-viewer path. Wrap in `htmltools::tagList()` so it routes through `print.shiny.tag.list`. For widgets with sizing sensitivity (d3heatmap) keep the call as a top-level chunk expression so `knit_print.htmlwidget` dispatches with the natural sizing policy.
- **Pre-register deps**: the touch-js-libs chunk near the top of the qmd touches each widget type once so its JS/CSS dependencies are registered in the document head. New widget types need a touch there.

## Test plan

- [ ] Snapshot tests pass on CI (the report HTML is in the snapshot)
- [ ] Eyeball the rendered before/after HTML side-by-side
- [ ] Spot-check that `disable_report_modules` still works (the `module_disabled()` helper retains its original semantics)
- [ ] Confirm the three intentional behaviour changes are acceptable

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pinin4fjords:
Before: 
[differentialabundance_report_BEFORE.zip](https://github.com/user-attachments/files/27726673/differentialabundance_report_BEFORE.zip)


After: 
[differentialabundance_report_AFTER.zip](https://github.com/user-attachments/files/27726684/differentialabundance_report_AFTER.zip)

atrigila:
May also close: https://github.com/nf-core/differentialabundance/issues/599

pinin4fjords:
> May also close: #599

@atrigila  I took a look and added one or more things. We've addressed as much of that issue as is sensible (left a comment there).

atrigila:
Noticed a few things while attemptng to do review:

1. The pipeline does not run with latest Nextflow version 26.04.1.
2. I downgraded and tried running the `test_full` with `NXF_VER=25.10.3 nextflow run . -profile test_full,singularity--outdir test` and got:
```
Validation failed for paramsheet row: deseq2_rnaseq_gsea,deseq2_rnaseq_gprofiler2. Error: null
/igenomes_base (s3://ngi-igenomes/igenomes/): could not validate file format of 's3://ngi-igenomes/igenomes/': Access Denied (Service: S3, Status Code: 403, Request ID: 8QSBDRNM0CFSCA2J, Extended Request ID: UZtwmYG1ltklagO+DNovWgBoPb7od7pJsaLNLZ16ZAr9XdTshMl1+eXKjqmDIoWY/2EwkOjZuJM=) (SDK Attempt Count: 1)
({"exploratory_final_assay":"variance_stabilised","custom_config_base":"http...d":"ls","differential_fc_column":"log2FoldChange","deseq2_shrink_lfc":true}): Value does not match against the schemas at indexes [20]
 -- Check script 'subworkflows/local/utils_nfcore_differentialabundance_pipeline/main.nf' at line: 371 or see '.nextflow.log' file for more details
``` 
The same happens with the `profile test`.
3. (This is not related to this PR) There are no examples in the documentation on how to run a 'simple' test using test data. This is necessary as the simple `nextflow run nf-core/differentialabundance -profile test,singularity--outdir test` won't work. 

I noticed that Claude says here:
> End-to-end on nf-dev-differentialabundance with nextflow run -profile test,docker against NXF_VER=25.10.3 - pipeline completes successfully.

but I cannot reproduce it using the same command because I get the above error.

pinin4fjords:
Lemme check...

assets/differentialabundance_report.qmd
```diff
@@ -48,1900 +45,1585 @@ library(UpSetR)
...
-            all_enriched,
-            caption_text
-          )
-        )
-      }
+### Differential `{r} p$features_type` details
 
-      cat("\n")
-    }
+```{r de-details, echo=FALSE, results='asis', warning=FALSE, message=FALSE}
```
atrigila, line 1530: While we have `warning=FALSE` here, with the nf-test "Test rnaseq dream profile" quarto results I see:

Image

assets/differentialabundance_report.qmd
```diff
@@ -48,1900 +45,1585 @@ library(UpSetR)
...
+    if (file.exists(model_file)) readLines(model_file, n = 1) else NA_character_
+})
 
-    p_value_types <- list(Adjusted = params$meta$params$differential_qval_column, Unadjusted = params$meta$params$differential_pval_column)
-    p_value_thresholds <- list(Adjusted = params$meta$params$differential_max_qval, Unadjusted = params$meta$params$differential_max_pval)
+captioned_table(
+    contrasts_to_print,
+    caption = "Table of contrasts",
+    pagination = nrow(contrasts_to_print) > 10,
+    full_width = FALSE,
```
atrigila: Perhaps this one could be `full_width` so that the contrasts are seen without scrolling:

Image

assets/differentialabundance_report.qmd
```diff
@@ -48,1900 +45,1585 @@ library(UpSetR)
...
-        !is.null(params$meta$params$features_name_col) &&
-        (params$meta$params$features_name_col %in% colnames(full_de))) {
-        label_col <- params$meta$params$features_name_col
-    } else {
-        label_col <- params$meta$params$differential_feature_id_column
-    }
+de_tool <- p$differential_method
+contrasts_to_print$model <- map_chr(contrasts_to_print$Id, \(id) {
+    model_file <- paste0(id, ".", de_tool, ".model.txt")
+    if (file.exists(model_file)) readLines(model_file, n = 1) else NA_character_
```
atrigila: We can remove the `model` column from `model_file` since it is the same as the formula column. 
We can replace `null` for empty as is in the other columns:

Image

assets/differentialabundance_report.qmd
atrigila: Perhaps we should output a more informative message here?

assets/differentialabundance_report.qmd
```diff
@@ -48,1900 +45,1585 @@ library(UpSetR)
...
+}
 
-htmltools::tagList(
-  rt,
-  htmltools::tags$div(
-    caption_text,
-    style = "text-align: center; margin-top: 10px; color: #5a677a;"
-  )
-)
+render_biotype_plot <- function(combined) {
```
atrigila, line 898: This is a lot better. Thank you.

assets/differentialabundance_report.qmd
```diff
@@ -48,1900 +45,1585 @@ library(UpSetR)
...
 
-    print(
-        htmltools::tagList(
-            htmltools::tags$p(
-                caption_text,
-                style = "font-weight: bold; margin-bottom: 6px;"
-            ),
-            rt
+    if (features_provided) {
+        merged <- merge(features, diff,
```
atrigila: Minor: I guess these base R `merge` could be tidyversed with joins such as:
```
merged <- features |>
    inner_join(
        diff,
        by = setNames(
            p$differential_feature_id_column,
            p$features_id_col
        )
    )
```

assets/differentialabundance_report.qmd
```diff
@@ -48,1900 +45,1585 @@ library(UpSetR)
...
-generate_boxplots <- function(combined, assay_data, observations, params,
-                               contrast_variables, main_grouping_variable,
-                               groupColorScale, label_col) {
-    qval_pretty <- prettifyVariablename(params$meta$params$differential_qval_column)
+# ---- Differential boxplots (top genes) -------------------------------------
 
-    # choose ID + display label columns in 'combined' after prettify
-    features_id_pretty <- prettifyVariablename(params$meta$params$features_id_col)
+generate_boxplots <- function(combined, label_col) {
+    qval_pretty <- prettifyVariablename(p$differential_qval_column)
```
atrigila, line 188: I've never used this package but I wonder if using something like https://memoise.r-lib.org/ could help speed up computation of some repeated stuff like when we use `prettifyVariablename` if a previously seen set of inputs is seen, it can return the previously computed output.

atrigila (commented):
The qmd looks a lot better with the refactoring. I like the tidyversing and the helper consolidation. I am leaving a few general comments while I try to run with full data.

pinin4fjords:
Thanks @atrigila, all three reproduce and all three are now sorted - just rebased and force-pushed.

### 1. Doesn't run with Nextflow 26.04.1

Already fixed on `dev` via the nf-core/tools v4.0.2 template merge (#707), the PR branch just hadn't picked it up. Verified on the test VM after rebase with `NXF_VER=26.04.1 nextflow run . -profile test,docker`: completes in 2m 6s, 25/25 tasks.

### 2. Schema validation `Error: null` + `Access Denied` on `s3://ngi-igenomes/igenomes/`

Root cause: nf-schema 2.5+ resolves `format: directory-path` defaults against the storage backend at launch, so the `igenomes_base` default `s3://ngi-igenomes/igenomes/` becomes a real S3 HeadBucket / ListBucket call. Environments that can't authorize that call (no creds, restricted SSO/IAM, on-prem) get the 403 you saw. The leading `Error: null` is the cosmetic `e.message`-is-null bug tracked in #718.

The fix has two halves and nf-core/rnaseq has both (see [rnaseq#1696](https://github.com/nf-core/rnaseq/pull/1696), [rnaseq#1739](https://github.com/nf-core/rnaseq/pull/1739), and the equivalent sarek work in [sarek#2184](https://github.com/nf-core/sarek/pull/2184)):

1. Drop `format: "directory-path"` from `igenomes_base` so the existence probe never runs. Already applied on `dev` via the template merge above.
2. Add `igenomes_base` to `validation.defaultIgnoreParams` in `nextflow.config` for belt-and-braces. Sneaked into this PR as a trivial one-liner along with a CHANGELOG entry.

Verified on the test VM with `NXF_VER=25.10.4 nextflow run . -profile test,docker --igenomes_base s3://nonexistent-bucket-xyz-12345/` (i.e. forcing the path to be unreachable): pipeline completes successfully in 2m 4s, 25/25 tasks.

### 3. No "simple test" docs example

Fair. Added a "Quick start with the test profile" subsection at the top of `docs/usage.md` with the canonical command. Trivial enough that I tucked it into this PR rather than spinning up a separate doc PR.

```bash
nextflow run nf-core/differentialabundance -profile test,docker --outdir test_results
```

---

Branch is rebased on current `dev` (force-pushed). Should be reviewable now without environment-specific blockers - sorry for the friction earlier.

assets/differentialabundance_report.qmd
pinin4fjords: These aren't R warnings - the chunk's `warning=FALSE` only suppresses `warning()` output. The blocks you saw were `cat()`-ed HTML from the per-contrast merge logic. In c58b5b46 I've consolidated both blocks into one informative `Note:` that explains the dropped IDs are expected (annotation source doesn't cover every matrix ID) and the on-disk differential table is unaffected. Kept the rare duplicate-IDs case as a real `Warning:` with annotation-deduplication guidance.

assets/differentialabundance_report.qmd
pinin4fjords: Flipped to `full_width = TRUE` in c58b5b46.

assets/differentialabundance_report.qmd
pinin4fjords: Dropped the model-file column in c58b5b46. The `null` strings turned out to be the literal text `"null"` that `read_metadata` writes for empty cells alongside `NA`, so the contrasts table now applies `replace_na(na_if(x, "null"), "")` across character columns. Verified on the regenerated dream report - no `"null"` cells.

assets/differentialabundance_report.qmd
pinin4fjords: 🙏

assets/differentialabundance_report.qmd
pinin4fjords: Swapped to `inner_join` with your `set_names` shape in c58b5b46.

assets/differentialabundance_report.qmd
pinin4fjords: Leaving this for a separate follow-up - adding a new package dependency for caching is out of scope for the refactor, and `prettifyVariablename` isn't a measured hotspot yet. Happy to open an issue if later profiling shows it matters.

atrigila (approved):
Looks good and works with the `test_full`. Thank you!

pinin4fjords:
Thanks for the review @atrigila !

PR #522, opened 2025-09-26
Replace `RMARKDOWNNOTEBOOK` with `QUARTONOTEBOOK`

delfiterradas:
Closes: https://github.com/nf-core/differentialabundance/issues/485
Closes: https://github.com/nf-core/differentialabundance/issues/486
Closes: https://github.com/nf-core/differentialabundance/issues/88

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/differentialabundance/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/differentialabundance _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core lint`).
- [ ] Ensure the test suite passes (`nf-test test main.nf.test -profile test,docker`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

PR #747, opened 2026-07-15
Fix/shinyngs contrast ordering

jozsefkun:
# Fix shinyngs contrast/result ordering mismatch causing wrong FC and p-values in Shiny app

_The description below was drafted with heavy AI assistance and is long/verbose; folded here to keep the PR page scannable (fold added during review, original text unchanged)._

<details>
<summary>Original PR description</summary>

## PR checklist

- [ ] If you've added a new tool - have you followed the pipeline conventions in the contribution docs?
- [ ] If necessary, also make a PR on the nf-core/differentialabundance branch on the nf-core/test-datasets repository.

---

## Description

I used Claude for debugging and writing the PR. Tested the changes by running the pipeline with the modified code.

### Bug

The shinyngs Shiny app silently displays fold-changes and p-values under the **wrong contrast label** when multiple contrasts are run. This is a data integrity issue: a user looking at contrast A in the app may actually be seeing the statistics for contrast B, with no warning or error.

### Root cause

The pipeline constructs the shinyngs `contrasts.csv` and the ordered list of differential result files from **two independent operations** on the same channel, and then re-sorts the CSV rows with `.sort().reverse()`:

```groovy
// ch_contrasts_sorted_shinyngs  (lines 768-775 in v2.0.0)
def content = contrast_map.collect { it.values().collect { val ->
        val ? val.toString() : ''
    }.join(',') }.sort().reverse()   // ← re-sorts CSV rows alphabetically descending
```

Meanwhile, `differential_results` — passed directly to `SHINYNGS_APP` — retains the original Nextflow channel emission order, which is **non-deterministic** (parallel DESEQ2/LIMMA jobs finish in unpredictable order).

The result: the CSV rows and the file list end up in different orders whenever the alphabetical-descending sort of contrast label strings does not happen to match the job completion order. The shinyngs `compile_contrast_data()` function in `R/accessory.R` then pairs them positionally (`V1`, `V2`, `V3`…), so mismatched positions silently assign the wrong statistics to each contrast.

The same bug exists in `ch_contrasts_sorted` (line 740), used for the HTML report contrast file.

### Fix

Sort `meta_with_contrast`, `results`, and `contrast_maps` together by `contrast.id` **before `multiMap`**, at the point where all three lists are still in guaranteed lockstep (having come from the same `groupTuple()`). Then remove `.sort().reverse()` from both `collectFile` closures — the order is correct by construction and deterministic.

By sorting upstream, inside the single `.map {}` closure where all three lists are available together, we avoid any possibility of the CSV and the file list diverging.

---

## Code change

In `workflows/differentialabundance.nf`, replace the two affected blocks as follows.

### Block 1: `differential_with_contrast` (lines ~720–734)

Apply the sort here so that both `ch_contrasts_sorted` and its `differential_results` sub-channel inherit the correct order.

**Before:**
```groovy
    differential_with_contrast = ch_paramsets
        .join( ch_differential_results
            .groupTuple()
        )
        .join( ch_contrasts )
        .map { meta, meta_with_contrast, results, contrast, variable, reference, target, formula, comparison ->
            def paramset_contrast_keys = contrast[0].keySet()
            def contrast_maps = meta_with_contrast.collect { it.subMap(paramset_contrast_keys) }
            [meta, meta_with_contrast, results, contrast_maps]
        }
        .multiMap { meta, meta_with_contrast, results, contrast_maps ->
            differential_results: [meta, meta_with_contrast, results]
            contrast_maps: [meta, contrast_maps]
        }
```

**After:**
```groovy
    differential_with_contrast = ch_paramsets
        .join( ch_differential_results
            .groupTuple()
        )
        .join( ch_contrasts )
        .map { meta, meta_with_contrast, results, contrast, variable, reference, target, formula, comparison ->
            def paramset_contrast_keys = contrast[0].keySet()
            def contrast_maps = meta_with_contrast.collect { it.subMap(paramset_contrast_keys) }
            // Sort meta_with_contrast, results and contrast_maps together by contrast id
            // so that the CSV rows and the result file list are always in the same
            // deterministic order, regardless of parallel job completion timing.
            def sorted = [meta_with_contrast, results, contrast_maps]
                .transpose()
                .sort { a -> a[0].id }
                .transpose()
            [meta, sorted[0], sorted[1], sorted[2]]
        }
        .multiMap { meta, meta_with_contrast, results, contrast_maps ->
            differential_results: [meta, meta_with_contrast, results]
            contrast_maps: [meta, contrast_maps]
        }
```

### Block 2: `ch_contrasts_sorted` collectFile (lines ~737–749)

Remove `.sort().reverse()` — order is now guaranteed by the upstream sort.

**Before:**
```groovy
    ch_contrasts_sorted = differential_with_contrast.contrast_maps
        .collectFile { meta, contrast_map ->
            def header = contrast_map[0].keySet().join(',')
            def content = contrast_map.collect { it.values().join(',') }.sort().reverse()
            def lines = header + '\n' + content.join('\n') + '\n'
            ["${meta.paramset_name}.csv", lines]
        }
```

**After:**
```groovy
    ch_contrasts_sorted = differential_with_contrast.contrast_maps
        .collectFile { meta, contrast_map ->
            def header = contrast_map[0].keySet().join(',')
            def content = contrast_map.collect { it.values().join(',') }
            def lines = header + '\n' + content.join('\n') + '\n'
            ["${meta.paramset_name}.csv", lines]
        }
```

### Block 3: `differential_with_contrast_shinyngs` (lines ~752–765)

Apply the same sort here for the shinyngs-specific channel (which is rebuilt after filtering formula-based contrasts).

**Before:**
```groovy
    differential_with_contrast_shinyngs = differential_with_contrast.differential_results
        .transpose()
        .filter { meta, contrast, results -> contrast.variable?.trim() }
        .groupTuple()
        .join( ch_contrasts )
        .map { meta, meta_with_contrast, results, contrast, variable, reference, target, formula, comparison ->
            def paramset_contrast_keys = contrast[0].keySet()
            def contrast_maps = meta_with_contrast.collect { it.subMap(paramset_contrast_keys) }
            [meta, meta_with_contrast, results, contrast_maps]
        }
        .multiMap { meta, meta_with_contrast, results, contrast_maps ->
            differential_results: [meta, meta_with_contrast, results]
            contrast_maps: [meta, contrast_maps]
        }
```

**After:**
```groovy
    differential_with_contrast_shinyngs = differential_with_contrast.differential_results
        .transpose()
        .filter { meta, contrast, results -> contrast.variable?.trim() }
        .groupTuple()
        .join( ch_contrasts )
        .map { meta, meta_with_contrast, results, contrast, variable, reference, target, formula, comparison ->
            def paramset_contrast_keys = contrast[0].keySet()
            def contrast_maps = meta_with_contrast.collect { it.subMap(paramset_contrast_keys) }
            // Sort all three lists together by contrast id before multiMap
            // so CSV rows and result files stay in the same deterministic order.
            def sorted = [meta_with_contrast, results, contrast_maps]
                .transpose()
                .sort { a -> a[0].id }
                .transpose()
            [meta, sorted[0], sorted[1], sorted[2]]
        }
        .multiMap { meta, meta_with_contrast, results, contrast_maps ->
            differential_results: [meta, meta_with_contrast, results]
            contrast_maps: [meta, contrast_maps]
        }
```

### Block 4: `ch_contrasts_sorted_shinyngs` collectFile (lines ~768–776)

Remove `.sort().reverse()` - order is now guaranteed by the upstream sort.

**Before:**
```groovy
    ch_contrasts_sorted_shinyngs = differential_with_contrast_shinyngs.contrast_maps
        .collectFile { meta, contrast_map ->
            def header = contrast_map[0].keySet().join(',')
            def content = contrast_map.collect { it.values().collect { val ->
                    val ? val.toString() : ''
                }.join(',') }.sort().reverse()
            def lines = header + '\n' + content.join('\n') + '\n'
            ["${meta.paramset_name}_shinyngs.csv", lines]
        }
```

**After:**
```groovy
    ch_contrasts_sorted_shinyngs = differential_with_contrast_shinyngs.contrast_maps
        .collectFile { meta, contrast_map ->
            def header = contrast_map[0].keySet().join(',')
            def content = contrast_map.collect { it.values().collect { val ->
                    val ? val.toString() : ''
                }.join(',') }
            def lines = header + '\n' + content.join('\n') + '\n'
            ["${meta.paramset_name}_shinyngs.csv", lines]
        }
```

### Block 5: Only for logging purposes, added a line immediately after `ch_contrasts` is defined (after line 432):

This prints the id used for sorting into the Nextflow log file.

**Before:**
```groovy
     ch_contrasts = ch_validated_contrast
        .splitCsv ( header:true, sep:'\t' )
        .map{meta, contrast ->
            contrast.blocking = contrast.blocking.replaceAll('^NA$', '')
            if (!contrast.id){
                contrast.id = contrast.values().join('_')
            }
            contrast.formula = contrast.formula?.trim() ? contrast.formula.trim() : null
            contrast.make_contrasts_str = contrast.make_contrasts_str?.trim() ? contrast.make_contrasts_str.trim() : null

            return [meta, contrast, contrast.variable, contrast.reference, contrast.target, contrast.formula, contrast.make_contrasts_str]
        }
        .groupTuple() // [meta, [contrast], [variable], [reference], [target], [formula], [comparison]]
```

**After:**
```groovy
    ch_contrasts = ch_validated_contrast
        .splitCsv ( header:true, sep:'\t' )
        .map{meta, contrast ->
            contrast.blocking = contrast.blocking.replaceAll('^NA$', '')
            if (!contrast.id){
                contrast.id = contrast.values().join('_')
            }
            contrast.formula = contrast.formula?.trim() ? contrast.formula.trim() : null
            contrast.make_contrasts_str = contrast.make_contrasts_str?.trim() ? contrast.make_contrasts_str.trim() : null

            return [meta, contrast, contrast.variable, contrast.reference, contrast.target, contrast.formula, contrast.make_contrasts_str]
        }
        .groupTuple() // [meta, [contrast], [variable], [reference], [target], [formula], [comparison]]

ch_contrasts.view { meta, contrast, variable, reference, target, formula, comparison ->
    "CONTRAST id=${contrast[0].id} variable=${variable[0]} reference=${reference[0]} target=${target[0]}"
    }
```
---

## Why `.sort().reverse()` was wrong

The comment above the block says:

> *"Save temporary contrast csv files with the entries **ordered by the differential results**"*

The intent was clearly to match the CSV row order to the file list order. But `.sort().reverse()` sorts CSV rows by their **string content** (alphabetical descending), completely independent of the `differential_results` list order. This defeats the stated goal.

The original non-determinism (parallel jobs finishing in unpredictable order) meant that neither the CSV rows nor the file list had a reliable order - but at least before `.sort().reverse()` was introduced they were both derived from the same `groupTuple()` emission and therefore matched each other. The `.sort().reverse()` broke that alignment.

The correct solution is to impose a **single deterministic sort key** (`contrast.id`) on all three lists together before they are split by `multiMap`.

---

## Impact

Any run of the pipeline with ≥2 contrasts where the shinyngs app is built (`shinyngs_build_app = true`) is potentially affected. The bug is silent - the app loads without error, but fold-changes and p-values in the differential results table are displayed under the wrong contrast labels.

</details>

---

## Update

The mechanism above was refined in a follow-up commit (see PR comments for detail):

- Order is established via `groupTuple(sort: { a, b -> a[0].id <=> b[0].id })` at the two grouping points, rather than a single explicit three-way sort. `groupTuple`'s `sort:` option sorts each of its output lists independently, so the paired values (e.g. `[meta, results]`) are nested as one element before grouping so the sort can't pull them apart.
- The contrast CSV text is now built inline, immediately after the order is fixed, instead of in a separate `collectFile` closure - removing the separate artifact that could be edited out of sync with the results list in future (the original regression's exact failure mode).
- The debug `.view()` call described above (Block 5) was removed before merge.
- Null-handling for CSV values is now unified across both paths (`val != null ? val.toString() : ''`), a small behavioral change from the previous code, which used a falsy check (`val ?`) on one path only - this also preserves `0`/`false` values that would previously have been blanked.


pinin4fjords:
Pushed a follow-up commit on top of this fix.

**What it does:** keeps the id-based contrast/result fix, but tightens where the ordering guarantee actually lives. `groupTuple(sort:)` sorts each of its output lists independently, not jointly, so meta and results are nested as one element before grouping so the sort can't pull them apart. The CSV text is now built inline, right where that order is already fixed, instead of in a separate `collectFile` closure that a later change could edit without noticing it needed to stay in sync with the results list. Also dropped a leftover debug `.view()` call.

**Where it broke:** the CSV and the results file list used to be two projections of the same channel (pre-2025), which kept them in lockstep for free. That changed with #462 (@atrigila, 2025-07-09), which split them into two independently-derivable lists to support formula-based contrasts. That didn't break anything by itself, but #522 (@delfiterradas, 2025-10-06, commit `9c70ce04`, "Sort the contrasts table") added `.sort().reverse()` to only the CSV side, for cosmetic ordering in the report, without realizing the results list depended on staying in the same order. That's the actual regression, and it shipped in v2.0.0.

pinin4fjords (approved):

pinin4fjords:
Thanks @jozsefkun , I corrected things a little but merging now (I want to add some shinyngs-related improvements soon)

